PR #2, opened 2016-07-15
Add provisioning scripts to repo.

rtnpro:
This migrates provisioning scripts from https://github.com/rtnpro/centos-cccp-ansible to this repo.


bamachrn:
This code base is not the latest code from centos-cccp-ansible repo so many thing are missing. We should pull the latest changes before pushing to this repo

PR #3, opened 2016-07-18
Changes for notification redirection to notification tube

bamachrn:


bamachrn:
Need to re-open the PR after https://github.com/CentOS/container-pipeline-service/pull/4 is merged as few variables changed in both PR. This will create merge conflict in either of these two.

PR #4, opened 2016-07-18
Allinone setup

rtnpro:
Usage:

```
ALLINONE=1 vagrant up
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,11 +13,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :hostmanager
   config.hostmanager.manage_host = true
   config.hostmanager.include_offline = true
-  config.ssh.insert_key = true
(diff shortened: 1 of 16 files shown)


bamachrn:
There is some problem with rsync_ssh_opts parameter. It not included in quotes (may be typo/intentional). When it is kept in quotes it is throwing error of variable undefined.

container-pipeline-service/provisions/roles/jenkins/slave/tasks/setup_worker.yml

rtnpro:
On Tue, Jul 19, 2016 at 1:37 PM, Bamacharan Kundu
<notifications@github.com> wrote:
>
> There is some problem with rsync_ssh_opts parameter. It not included in quotes (may be typo/intentional). When it is kept in quotes it is throwing error of variable undefined.
>
>
I fixed that issue. Can you check now?

bamachrn:
This looks good now +1 for merging

PR #5, opened 2016-07-19
Added new service scripts for worker.

rtnpro:
This adds new service scripts for beanstalk workers:

- dispatcher worker
- delivery worker


bamachrn:
looks good for automating start of worker services. we need to add restart logic later on this.

rtnpro:
It's not for the restart logic, but to add the service files to consume in the allinone vagrant box setup.

PR #6, opened 2016-07-20
read the duffy key from the workspace on the slave instead of checking it in directly

bstinsonmhk:
The CentOS CI Service provides the duffy key in ~/duffy.key, we should read that in directly instead of checking it in to version control.

This PR will need a JJB run to update the jobs.


kbsingh:
@rtnpro comments?

rtnpro:
:+1:

PR #7, opened 2016-07-20
Add atomic scan based image scanner

dharmit:
@kbsingh @bamachrn ping

### DIFF
--- a/pipeline_scanner/Dockerfile
+++ b/pipeline_scanner/Dockerfile
@@ -1,0 +1,10 @@
+FROM centos:centos7
+
+LABEL INSTALL='docker run -it --rm --privileged -v /etc/atomic.d:/host/etc/atomic.d/ $IMAGE sh /install.sh'
+
+# Install python-docker-py to spin up container using scan script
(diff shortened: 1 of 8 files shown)


dharmit:
As far as migrating existing scanner to use `atomic scan` based scanner is concerned, I've not replaced the code in `beanstalk_worker` directory yet. This is to ensure we can do some testing (which I've already done and faced no issues on Devcloud) before we push it further.

Let me know your thoughts.

kbsingh:
Can we use the existing beanstalkc.py in the repo rather than carry another copy here ?

bamachrn:
This looks good to me. We need to figure out some way to use same beanstalkc.py for the workers, this is the reason I kept all the workers in the same directory. Scan runs separately and keeping scaling in mind this looks good.

dharmit:
This worker should eventually replace the one in `beanstalk_worker` directory. Once that's done, we won't need to carry the second copy of `beanstalkc.py`.

kbsingh:
Lgtm

PR #11, opened 2016-07-26
Fix vagrant allinone setup

rtnpro:


Vagrantfile
```diff
@@ -66,6 +58,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       inventory_path = "provisions/hosts.vagrant"
   end
 
+  # Ensure Jenkins SSH keys exist
+  system('if [ ! -f /tmp/cccp-jenkins.key ]; then ssh-keygen -t rsa -N "" -f /tmp/cccp-jenkins.key; fi')
+
   config.vm.define "master" do |master|
     master.vm.hostname = "cccp"
     master.vm.network :private_network, ip: "192.168.100.100"
```
navidshaikh: IMO, IP address should be defined in a constant and reused.  I have seen cases whereI create multiple VMs and the IP address conflicts.

Defined in a variable like `VM_IP = "192.168.100.200` atop Vagrantfile, the variable can be re-used at all places like "#{VM_IP}".

"PUBLIC_ADDRESS" variable name is used for such use cases in ADB vagrantfiles.

navidshaikh: Although, the review is not on the code change part of PR, if can improve it as well.

dharmit:
@rtnpro Seeing error with this PR when I do `ansible-playbook -i hosts -s vagrant.yml -vvvv`. [Pastebin](http://pastebin.centos.org/50831/).

kbsingh:
I've run through a ALLINONE setup, and was left with a working single box installation.
LGTM; Merging

PR #20, opened 2016-08-10
Added logs and try catch for job handling

bamachrn:
* All the logs for worker is written to stdout, which would be available from journalctl
* Try catch blocks are added so that job only gets deleted when task is complete.


kbsingh:
LGTM

PR #21, opened 2016-08-12
Schedule scan fix

dharmit:
@bamachrn ping


kbsingh:
lgtm,

we still need to get some work done around getting the schedular up.

PR #23, opened 2016-08-12
Changes for getting worker files running as service

bamachrn:
*Added functions to build and delivery worker for log and command
* Added kube config to run all the commands independently in build and delivery worker
* Changed client/build.sh and added kubeconfig to login to openshift
* Changed provisioning scripts to get node.kubeconfig to client and workers


kbsingh:
lgtm

Issue #16, opened 2016-08-05
Add ability for user to specify custom tag, other than latest for their container builds.


mohammedzee1000:
We should work on getting the pipeline service to be able to tag images, other than latest. Preferably, we allow the user to specify their image tag in the cccp.yml which should allow them to control it from their repository.

PR #30, opened 2016-08-18
Added desired-tag to index.yml

bamachrn:
* Now we are taking one more parameter as desired-tag in index.yml
* This is the final tag of the image pushed to public registry
* If no desired-tag is provided it will default to latest


kbsingh:
@mohammedzee1000 we will need to edit the readme to include this as well, can you propose a PR for that addition in the docs side ?

mohammedzee1000:
@kbsingh will do

bamachrn:
Fixes https://github.com/CentOS/container-pipeline-service/issues/16

PR #31, opened 2016-08-19
Added script to transform index.d to index.yml and vice versa.

mohammedzee1000:
This in in lieu with the decision to move from single index.yml to index.d directory with multiple ymls which will be combined into index.yml on the fly by the pipeline so no major changes will be required from the code perspective.


kbsingh:
Do you want to include this changeset in the PR that uses this in the code base ?

mohammedzee1000:
Changes done to the index reader script. Now instead of passing index.yml path, you pass the path of the index.d directory, the 2 extra lines i have added will generate index.yml in pwd and the rest of the process will continue as usual

 Working on testing this in all in one

PR #33, opened 2016-08-22
Added script to transform index.d to index.yml and vice versa.

mohammedzee1000:
Changes verified on allinone.

Will require modification of container index to the new index.d method before merging.


jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,199 @@
...
+            else:
+
+                desiredtag = project["desired-tag"]
+
+            if appid not in data.keys():
+
+                data[appid] = []
+
+            data[appid].append({
+
```
rtnpro: A new line here is not needed

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,199 @@
...
+                "job-id": jobid,
+                "git-url": giturl,
+                "git-path": gitpath,
+                "git-branch": gitbranch,
+                "target-file": targetfile,
+                "notify-email": notifyemail,
+                "desired-tag": desiredtag,
+                "depends-on": dependson
+
```
rtnpro: same here! no newline needed

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,199 @@
...
+                    os.remove(filenm)
+
+                with open(filenm, "a+") as thefile:
+
+                    safe_dump(value, thefile, default_flow_style=False)
+
+        return
+
+    def run(self):
+
```
rtnpro: no new line needed

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,199 @@
...
+        if not self._reverse:
+
+            potentialfiles = self._indexdotd + "/*.yml"
+            self._genindexyml(glob(potentialfiles))
+            return self._indexyml
+
+        self._genindexd()
+        return self._indexdotd
+
+def initparser():
```
rtnpro: if a name is a multi word name, use ``_`` to separate that!

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,199 @@
...
+import sys
+from argparse import ArgumentParser
+
+
+class IndexGenerator:
+
+    def __init__(self, indexdpath="./index.d", indexyml="./index.yml", reverse=False):
+
+
+        #if not os.path.isabs(indexdpath):
```
rtnpro: Don't keep unused code like this dangling in the codebase

rtnpro:
@mohammedzee1000 

The could use some cleanup and enhancements to make it easy to review:

- stop putting unnecessary newlines
- split multi word variable names with ``_``
- the code could use a lot of inline comments

Make these changes and ping me to review again.

mohammedzee1000:
@rtnpro changes done

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,196 @@
...
+                namespace = file_name.split(".")[0]
+
+            # Otherwise the project namespace is the entire filename
+            else:
+                namespace = file_name
+
+            potential_index_data = None
+
+            # Read the yaml data in the file
+            with open(potential_index_file, "r") as thefile:
```
rtnpro: why ``thefile``, why not just use ``f``?

mohammedzee1000: i dont like using x type variable at same time could'nt figure out a proper name.

mohammedzee1000:
@rtnpro fixed :)

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,196 @@
...
+                potential_index_data = load(thefile)
+
+            # Go through every entry in the yaml data file, extracting what is needed.
+            for entry in potential_index_data:
+                tid = current_index
```
rtnpro: why unnecessarily copy the dict data into vars, if you are not changing something? directly consume the values from the dict in line 73 onwards

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,196 @@
...
+        else:
+            if not os.path.exists(self._index_yml):
+                raise ValueError("Invalid path specified")
+
+        return
+
+    def _genindexyml(self, potential_index_files):
+        """Generate index.yml from index.d files"""
+
+        current_index = 1
```
rtnpro: There's no need to track index using ``current_index`` var, you can just use ``enumerate`` in line 39, and use ``n`` as as the index, unless you have a special reason to do so.

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,196 @@
...
+                })
+
+                current_index += 1
+
+        # Dump the indexyml data to specfied index.yml
+        with open(self._index_yml, "w") as index_yml_file: 
+            safe_dump(index_yml, index_yml_file, default_flow_style=False, indent=2)
+
+        # Add few lines of comment indicating the the file is generated.
+        cl1 = "This file is generated by the CentOS Community Container Pipeline Service"
```
rtnpro: Why use ``cl1`` and ``cl2``. You can just use a multiline string like:

```
"""
# foo bar
# blah
"""
```

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,196 @@
...
+        index_yml = None
+        indexd_data = {}
+
+        # Open the indexyml file and read the yaml data
+        with open(self._index_yml, "r") as index_file:
+            index_yml = load(index_file)
+
+        # For every project in the indexyml file, collect the nessasary data
+        for project in index_yml["Projects"]:
+            app_id = project["app-id"]
```
rtnpro: Same here, I do not see any point in copying the values here, as well.

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,196 @@
...
+            # If no desired-tag is specified, assume a default, otherwise take specified desired-tag
+            if "desired-tag" not in project.keys():
+                desired_tag = "latest"
+
+            else:
+                desired_tag = project["desired-tag"]
+
+            # If there isnt already indexd data entry for a specific appid, add it.
+            # The appid will be used later on to set the filenames
+            if app_id not in indexd_data.keys():
```
rtnpro: You don't need to do this check if ``indexed_data`` is a ``defaultdict``.

```
from collections import defaultdict

indexd_data = defaultdict('str')
```

rtnpro:
@mohammedzee1000 Go through my above comments and fix the issues I have raised, and we'll be good to merge.

jenkinsbuilder/indexdindexymltransform.py
mohammedzee1000: im using the current_index to keep a count to use as the "id" field in the index.yml file.

Basically, as i am reading through the index.d files, the first entry gets id: 1 and the second one gets id: 2 and so on. I dont think enumeration is what is required here.

rtnpro: If you say so, :+1:

jenkinsbuilder/indexdindexymltransform.py
mohammedzee1000: This is for the reverse process, where you have an index.yml but for some reason, want the index.d files.

jenkinsbuilder/indexdindexymltransform.py
mohammedzee1000: The indexd_data is used to store the content to be written into the index.yml file

For example if postgres and centos are 2 appids index.yml, then
indexd_data["postgres"] = all entries in index.yml of postgres app-id
indexd_data["centos"] = all entries in index.yml of centos app-id

Then i iterate through the dict creating a ${key}.yml file and dumping ${value} which are all entries into that file so

indexd_data.key = centos hence filename = "centos.yml" and data = "value" which is yml pertaining to centos containers.

jenkinsbuilder/indexdindexymltransform.py
mohammedzee1000: This is now fixes, will add to PR

jenkinsbuilder/indexdindexymltransform.py
rtnpro: So, you can have a ``defaultdict(list)`` or ``defaultdict(dict)`` as required. Your answer does not explain to me a reason of not using ``defaultdict``.

Also, ``if app_id not in indexd_data.keys()`` has ``O(n)`` complexity, whereas ``if app_id not in indexd_data`` has ``O(1)`` complexity.

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,197 @@
...
+        index_yml = None
+        indexd_data = defaultdict()
+
+        # Open the indexyml file and read the yaml data
+        with open(self._index_yml, "r") as index_file:
+            index_yml = load(index_file)
+
+        # For every project in the indexyml file, collect the necessary data
+        for project in index_yml["Projects"]:
+            app_id = project["app-id"]
```
rtnpro: Stop copying data unnecessarily into vars

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,182 @@
...
+
+                current_index += 1
+
+        # Dump the indexyml data to specfied index.yml
+        with open(self._index_yml, "w") as index_yml_file: 
+            safe_dump(index_yml, index_yml_file, default_flow_style=False, indent=2)
+
+        # Add few lines of comment indicating the the file is generated.
+        commentline = """This file is generated by the CentOS Community Container Pipeline Service\n
+                      Please do not edit this file manually.\n\n
```
rtnpro: This will lead to two ``\n``.

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,182 @@
...
+            # Dump the yml value to the key.yml in index.d
+            if key is not None:
+                file_name = self._index_dotd + "/" + key + ".yml"
+
+                if os.path.exists(file_name):
+                    os.remove(file_name)
+                with open(file_name, "a+") as the_file:
+                    safe_dump(value, the_file, default_flow_style=False)
+
+        return
```
rtnpro: A method always returns ``None`` by default. Remove this line.

jenkinsbuilder/indexdindexymltransform.py
```diff
@@ -0,0 +1,182 @@
...
+                      Please do not edit this file manually.\n\n
+                      """
+
+        with open(self._index_yml, "r") as original:
+            data = original.read()
+
+        with open(self._index_yml, "w") as modified:
+            modified.write(data)
+
+        return
```
rtnpro: No point of having this line as well.

mohammedzee1000:
@rtnpro done

jenkinsbuilder/indexdindexymltransform.py
rtnpro: @mohammedzee1000 ping?

mohammedzee1000: Yes, apparently one of the keys returned in the kv pair from indexd_data is None, which caused failure at this point. Thats the reason i put in that check for key not being null

mohammedzee1000: @rtnpro yea i thought so to, but apparently one of the keys was null so i was forced to put that check in.

rtnpro: @mohammedzee1000 

I don't see how is that gonna fix the problem. You are not expecting any output from ``_genindex`` anyways, so why bother to have the return statement. If there's a corner case somewhere else, fix that there.

mohammedzee1000: Hmm i though the return statements were compulsory in the function, never tried not having them so didnt know.

rtnpro:
LGTM :+1:

kbsingh:
I am not entirely sure why we need this much of code to achieve what is essentially a : "cat index.d/* > index.yml" - and then let the rest of the tooling just work as is (ie. the jenkins jjb side, the index validation script etc )

mohammedzee1000:
Validated in allinone box

kbsingh:
#dotests

rtnpro:
#dotests

mohammedzee1000:
Will setup new branch as per discussion

PR #40, opened 2016-08-24
Block all the user accessing through https for requests other than GET

bamachrn:
* Blocking all the users to do any operation other than pull from registry.centos.org
* Any one tries to do other operation, it returns 403


kbsingh:
lgtm

PR #52, opened 2016-08-31
add a job spec to run the validator script over the index 

bstinsonmhk:
Starting on a jobspec to get things running in cico. Once we get master of the index in a good state we can enable pr-level testing on this job.


kbsingh:
lgtm

PR #55, opened 2016-09-03
Cleaned up testindex code

mohammedzee1000:
The cleanup includes rewiring of code making it far simpler and leaner. The script no longer bothers about getting the index ready for testing, instead expecting user to provide path to index,d without worrying how it got there. Also, the tests happen directly on the index.d files instead of combining them, making it easier to isolate the entry which caused the problem. 

Also, overhauled the logging system into what is now a summary at the end of the entire run.


rtnpro:
#dotests

PR #57, opened 2016-09-06
Fix #48 #56 : Consolidates the results from multiple scanners

### LINKED ISSUE #48: Consolidate / validate the atomic scanner results
As part of the pipeline, one of the step is to scan built container image using the atomic scanner(s) we have. The results generated by atomic scanner need to be consolidated and presented. 

If atomic scanner is providing the scan result score, and based on the scan score / validation, we may need to notify the author about issues.

The aim of this issue is to answer following questions:
  - What do we do with the results of atomic scanners?
  - Decide, If there are results from scanner which should block the pipeline progress of image under process.
  - How do we present the results of atomic scanners?

### LINKED ISSUE #56: pipeline-scanner yum update verification bug
The results from pipeline-scanner are not verified from actual output of `yum` command, but rather checked for JSON output, which will be true in all the cases.

navidshaikh:
Fixes #56: Fixes verifying the results from pipeline scanner.
  Now checks for actual result field from scanner and sets the
  value for "msg" field accordingly.

  Fixes #48: This commit also consolidates results from pipeline
  scanner and scanner-rpm-verify and sends over tube in one go.

  The data structure of the info sent over the message bus has changed.
  The updates data structure for the output:

  ```
  d = {
  "image_under_test": <full_image_name>,
  "action": "completed_scheduled_scan",
  "namespace": <job_data.namespace>,
  "msg": {
    "scanner-rpm-verify": "<scanner_specific_message>",
    "pipeine-scanner": "<scanner_specific_message>"
  },
  "logs": {
    "scanner-rpm-verify": {<scanner_json_data>},
    "pipeline-scanner": {<scanner_json_data>}
  }
}
  ```


bamachrn:
#dotests

kbsingh:
#dotests

bamachrn:
Notify_user_worker should not be the point for checking scanner logs. Can we not go for another worker without altering the notify_worker? I believe that would keep things simpler and effective.

rtnpro:
#dotests

navidshaikh:
Following items to be added in this PR:

- [ ] More logging

- [ ] add systemd service for worker notifying user with scan results

- [ ] install the new scanner via ansible provisioning

beanstalk_worker/worker_start_scan.py
```diff
@@ -26,212 +26,340 @@
...
+                raise
+        else:
+            # TODO: do not exit here if one of the scanner failed to run,
+            # others might run
             logger.log(
                 level=logging.FATAL,
-                msg=str(error)
-                )
-            return
+                msg="Error running the scanner %s. Error: %s" % err)
```
dharmit: Shouldn't there be one more variable here? I think `self.scanner_name`.

beanstalk_worker/worker_start_scan.py
```diff
@@ -26,212 +26,340 @@
...
+                json_data = json.loads(open(output_json_file).read())
+            else:
+                logger.log(
+                    level=logging.FATAL,
+                    msg="No scan results found at %s" % output_json_file)
+                # FIXME: handle what happens in this case
+                raise
+        else:
+            # TODO: do not exit here if one of the scanner failed to run,
+            # others might run
```
dharmit: :+1: for the TODO. As we add more scanners, we need to figure how to go about handling such a scenario. I think a good way would be to add a message saying `xyz scanner failed`. And, as we start adding more scanners, doing that sooner rather than later would make sense.

beanstalk_worker/worker_start_scan.py
```diff
@@ -26,212 +26,340 @@
...
+            level=logging.INFO,
+            msg="Removing the image %s" % image_under_test)
+        conn.remove_image(image=image_under_test, force=True)
+
+    def run(self, image_under_test):
+        """
+        Run the scanner
+        """
+        self.mount_image(image_under_test)
+        image_id = self.atomic_object.get_input_id(image_under_test)
```
dharmit: `image_id` is being fetched in `mount_image` function as well. Maybe return it from that function to avoid repetition or maybe initialize `self.image_id` in `__init__` function with the value `None` and set it once when `run` method is executed?

beanstalk_worker/worker_start_scan.py
```diff
@@ -26,212 +26,340 @@
...
-            "name_space": namespace,
-            "notify_email": notify_email
-        }
-    bs.use("master_tube")
-    jid = bs.put(json.dumps(d))
-    logger.log(
-        level=logging.INFO,
-        msg="Put job on master tube with id: %d" % jid
-    )
+        if not json_data["Package Updates"]:
```
dharmit: I'm a little confused here about when this condition will hold `True`. Wouldn't it evaluate to `True` when `Package Updates` key holds no data? Which is contrary to what we're looking for.

```python
In [17]: d = {}

In [18]: d["updates"] = None

In [19]: d
Out[19]: {'updates': None}

In [21]: if d["updates"]:
    print "if"
else:
    print "else"
   ....: 
else

In [22]: if not d["updates"]:
    print "if"
else:
    print "else"

In [23]: d["updates"] = 123

In [24]: if not d["updates"]:
    print "if"
else:
    print "else"
   ....: 
else

```

navidshaikh:
> @navidshaikh suggested minor changes and one peculiar confusion with Package Updates part. Looks good otherwise.

I updated the PR addressing the minor changes, also fixed the package updates part, thanks pointing it out.

atomic_scanners/worker_start_test.py
```diff
@@ -1,213 +0,0 @@
...
-bs.watch("start_schedule_scan")
-
-while True:
-    try:
-        job = bs.reserve()
-        job_data = json.loads(job.body)
-        test_job_data(job_data)
-        job.delete()
-    except Exception as e:
-        logger.log(level=logging.FATAL, msg=e.message)
```
navidshaikh: This file was moved into `beanstalk_worker`  directory and it needed to be removed.

beanstalk_worker/worker_start_scan.py
```diff
@@ -26,212 +26,396 @@
...
 
-# set blank values so that they can be used later to clean environment
-
-image_id = ""
-image_rootfs_path = ""
+SCANNERS = {
+        "registry.centos.org/pipeline-images/pipeline-scanner": [
+            "image_scan_results.json"],
+        "registry.centos.org/pipeline-images/scanner-rpm-verify": [
+            "RPMVerfiy.json"]}
```
navidshaikh: Add a lookup hash per scanner and their output filenames.

mail_service/worker_report_scan_results.py
```diff
@@ -0,0 +1,78 @@
...
+  email_body = compose_email_text(job_details)
+
+  print "==> Sending scan results to user %s " % notify_email
+  command = "/mail_service/send_mail.sh"
+  subprocess.call([
+      command,
+      subject,
+      email_body,
+      notify_email,
+      ])
```
navidshaikh: After sending scan results to user, lets relay the job to next phase, i.e. delivery phase

navidshaikh:
The PR is code complete now, testing the PR for end to end flow.

navidshaikh:
> The PR is code complete now, testing the PR for end to end flow.

Tested the end to end flow, added a commit to improve the email contents for scanning results, rebased the PR on master.

beanstalk_worker/worker_start_scan.py
```diff
@@ -26,212 +26,424 @@
...
+
+    def run_a_scanner(self, scanner, image_under_test):
+        """
+        Run the given scanner on image_under_test
+        """
+        json_data = {}
+
+        # If scanner is pipeline-scanner then hand over to specific class,
+        # since pipeline-scanner mounts the image first and then executes
+        if scanner == "registry.centos.org/pipeline-images/pipeline-scanner":
```
rtnpro: This is unnecessary repetition of code across the scanners. You could do something like:

``` python
class ScannerRunner:
    # OrderedDict if order is important, else just a dict
    scanners = OrderedDict(
        {
            "registry.centos.org/pipeline-images/pipeline-scanner": PipelineScanner,
            "registry.centos.org/pipeline-images/scanner-rpm-verify": ScannerRpmVerify
         }
    )

    def run_scanner(self, name, image_under_test):
        scanner = self.scanners.get(name)()
        # then you could have the following code

    def run_scanners(self, image_under_tests):
        for name, mod in self.scanner.items():
            scanner = mod()
            # run scanners
```

beanstalk_worker/worker_start_scan.py
```diff
@@ -26,212 +26,424 @@
...
+                msg="Removing the image: %s" % image_under_test)
+
+        conn.remove_image(image=image_under_test, force=True)
+
+        # TODO: Check here if at least one scanner ran successfully
+        logger.log(level=logging.INFO, msg="Finished executing all scanners.")
+        return True, scanners_data
+
+
+class ScannerRPMVerify(object):
```
rtnpro: Also, the Scanners should have the same structure, which seems to be the case here. So, it'd be great if we create a base class for other scanners to inherit and extend from.

mohammedzee1000: I would second this.

navidshaikh: There are two types of scanners -
 - scanners needing image_under_test to be mounted (readwrite)
 - scanners not needing image_under_test to be mounted (readonly)

I am planning to structure it in following way
1 - A main scanner handler (base class)
2 - A class for (all) readonly scanners ( inherit from 1)
3 -  A class for (all) readwrite scanners (inherit from 1)
N - Individual scanners data validations [..] (this will later be transformed into configs)


The reason, I did not put everything in single PR, is to minimize the total code addition / deletion. At the moment, we have 2 scanners, this PR is scoped around that.

Beside that I agree on the suggestion. (y)

mohammedzee1000: Cool, just be careful you don't end up with the diamond problem of inheritance at some point.

rtnpro:
@navidshaikh https://github.com/CentOS/container-pipeline-service/pull/57/commits/cfac86cfba96dd630e4b98b1c04d98de99fa5c87 looks good. :+1:

dharmit:
LGTM

PR #61, opened 2016-09-08
Changed transformer and textindex to handle modification to index.d 

mohammedzee1000:
The index.d are now proper yml files. So minor changes were required
to remove the explicit corrections introduced by testindex and transformer
and also use yamls own librarires to generate the index.yml file.


kbsingh:
merging to pair with https://github.com/CentOS/container-index/pull/37

PR #67, opened 2016-09-09
Don't run the scanner if running in CI environment

dharmit:
ping @kbsingh @bamachrn @rtnpro

### DIFF
--- a/beanstalk_worker/worker_start_test.py
+++ b/beanstalk_worker/worker_start_test.py
@@ -208,7 +208,17 @@ while True:
     try:
         job = bs.reserve()
         job_data = json.loads(job.body)
-        test_job_data(job_data)
+        if len(sys.argv) > 1 and sys.argv[1] == "ci":
+            # If working in CI environment, don't run the scanner
+            logger.log(level=logging.INFO, msg="CI check...")
+            job_data["action"] = "start_delivery"
+            job_data["msg"] = ""
+            job_data["logs"] = ""
+            bs.use("master_tube")
+            jid = bs.put(json.dumps(job_data))
+        else:
+            test_job_data(job_data)
+
         job.delete()
     except Exception as e:
         logger.log(level=logging.FATAL, msg=e.message)


kbsingh:
looks ok to me, is there a corrosponding issue for the problem we are trying to fix, if not - can we make sure that there is an issue setup so we can clear this out in the coming weeks properly.

beanstalk_worker/worker_start_test.py
```diff
@@ -208,7 +208,17 @@ def test_job_data(job_data):
     try:
         job = bs.reserve()
         job_data = json.loads(job.body)
-        test_job_data(job_data)
+        if len(sys.argv) > 1 and sys.argv[1] == "ci":
+            # If working in CI environment, don't run the scanner
+            logger.log(level=logging.INFO, msg="CI check...")
+            job_data["action"] = "start_delivery"
+            job_data["msg"] = ""
+            job_data["logs"] = ""
```
dharmit: >logs are otherwise sent as JSON, if logs are empty, we should return {} instead of empty string.

You suggest we use `{}` instead of `""`? Ack, works for me, I can change it.

> Although both values can fail the if condition, we can be consistent in the type of value field for every key.

But if value for one key is a list and for other it's a string, how do we keep consistency in the overall json? Maybe I'm missing something here. How about we use `None` for empty values?

dharmit:
@kbsingh done

Fixes #69

beanstalk_worker/worker_start_test.py
navidshaikh: We can not inforce to have all values for all keys in JSON to be of same data types, however we can be consistent setting same data type value for *particular* key across the code.

dharmit: Okay. And afaik, I have instantiated both `msg` and `logs` with `""`. It's happening at two places. lines 215-216 and 58-59

beanstalk_worker/worker_start_test.py
```diff
@@ -208,7 +208,17 @@ def test_job_data(job_data):
...
+            bs.use("master_tube")
+            jid = bs.put(json.dumps(job_data))
+        else:
+            test_job_data(job_data)
+
```
navidshaikh: The code changes proposed in this PR is already part of master branch check https://github.com/CentOS/container-pipeline-service/blob/master/beanstalk_worker/worker_start_test.py#L211-L221 

and the purpose of the PR is achieved, i.e. not to run the scanner if executing in `ci` env.

PR #71, opened 2016-09-13
Complete end to end CI

rtnpro:


provisions/roles/scanner/tasks/main.yml
```diff
@@ -48,11 +52,20 @@
   with_items:
       - {src: "../beanstalk_worker/cccp-test-worker.service", dest: /etc/systemd/system/cccp-test-worker.service}
 
-- name: Enable Scanner worker
-  service: name="cccp-test-worker" state=restarted enabled=yes
+- name: Skip atomic scan for test environment
+  replace: >
+    dest=/etc/systemd/system/cccp-test-worker.service
+    regexp='worker_start_test.py$'
+    replace='worker_start_test.py ci'
```
bamachrn: This will require changes in the service file we have for test worker. So that it takes command line args

rtnpro: @bamachrn why? it's working as it is now.

bamachrn: Yeah, got it.

bamachrn:
LGTM :+1:

Issue #63, opened 2016-09-09
Error while placing build, test, delivery_scripts in proper place


bamachrn:
* For some cases when the users docker file changes its user to non superuser. Parsing of cccp.yml raises error, as it can not install PyYAML to container. As a result the expected build_script is not placed properly. As a result the build_script is showing not available.
* Also we are changing the dockerfile to add cccp.yml and cccp_reader in proper place to parse.

PR #76, opened 2016-09-30
Cccp reader fix

bamachrn:
* Removed yml parser from built image
* Added build,test, delivery scripts from cccp_reader and created links for running them
* For now disabled running build time scripts

### DIFF
--- a/server/Dockerfile.build
+++ b/server/Dockerfile.build
@@ -2,6 +2,7 @@ FROM centos
 
 RUN yum -y install docker
 RUN yum -y install git
+RUN yum install -y PyYAML libyaml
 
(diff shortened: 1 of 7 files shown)


server/Dockerfile.build
```diff
@@ -2,6 +2,7 @@ FROM centos
 
 RUN yum -y install docker
 RUN yum -y install git
+RUN yum install -y PyYAML libyaml
```
dharmit: Is it possible to install packages in one `RUN` instruction? Just to avoid additional layers.

bamachrn:
fixes https://github.com/CentOS/container-pipeline-service/issues/63

PR #77, opened 2016-09-30
Update CI Readme to include pointers to CI jobs.

rtnpro:


kbsingh:
fantastic, thanks for adding it to the docs, should be easier to find.

PR #81, opened 2016-10-03
Added checks for invalid characters and reformatted code.

mohammedzee1000:
This adds a check for presence of .,/ or _ in app-id and job-id
fields and also applied some code reformating.


mohammedzee1000:
@rtnpro @bamachrn can one of you do #dotests on this.

kbsingh:
ok to test

PR #82, opened 2016-10-03
Trigger Dockerfile linter

dharmit:


dharmit:
This PR aims to trigger the dockerfile lint checker before building the image. Need feedback @bamachrn @rtnpro

dharmit:
#dotests

dharmit:
#dotests

kbsingh:
ok to test

dharmit:
#dotests

dharmit:
#dotests

dharmit:
I don't understand why tests keeps throwing permission denied:

```
line 4: ./lint-dockerfile.sh: Permission denied", "Build step 'Execute shell' marked build as failure",
"Finished: FAILURE", "Completed bamachrn-python #1 : FAILURE"], "warnings": []}
```

It's got executable permissions as well as `#!/usr/bin/bash` at the start of the file like the other shell script in the same directory.

Need inputs @bamachrn @rtnpro.

kbsingh:
Ok to test

dharmit:
#dotests

dharmit:
#dotests

dharmit:
#dotests

mohammedzee1000:
LGTM

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -39,9 +39,10 @@ def lint_job_data(job_data):
     namespace = job_data.get('name_space')
 
     dockerfile = job_data.get("dockerfile")
+    to_email = job_data.get("to_email")
```
navidshaikh: No field through out the code now uses `to_email`, `notify_email` is the key you should use and access.

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -59,7 +60,8 @@ def lint_job_data(job_data):
     response = {
         "logs": out,
         "action": "notify_user",
-        "namespace": namespace
+        "namespace": namespace,
+        "to_email": to_email
```
navidshaikh: (as commented above)

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,34 @@
...
+import sys
+
+if __name__ == "__main__":
+    name_space = sys.argv[1]
+    git_path = sys.argv[2]
+    filename = sys.argv[3]
+    notify_email = sys.argv[4]
+
+    # having '/' in a value used in os.path.join generates unexpected paths
+    if git_path[0] == "/":
```
navidshaikh: use `startswith` on strings

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,34 @@
...
+        git_path = git_path[1:]
```
navidshaikh: use `startswith` on strings

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,34 @@
...
+
+    if filename[0] == "/":
```
navidshaikh: use `startswith` on strings

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,34 @@
...
+        filename = filename[1:]
+
+    lint_job_data = {}
+
+    with open(os.path.join(os.environ['DOCKERFILE_DIR'], git_path, filename)) \
+    as f:
+        dockerfile = f.read()
```
navidshaikh: We should do error handling here, there may be issues accessing environment variables

dharmit: I can imagine one issue which is that environment variable isn't set. But we're explicitly setting that [here](https://github.com/CentOS/container-pipeline-service/pull/82/files/2819e64342fa9fe9ab6fce9d8e054a60ded1cdc7#diff-4b6033ad5d043bfedd9cae16205e449cR15).

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -59,7 +60,8 @@ def lint_job_data(job_data):
     response = {
         "logs": out,
```
dharmit: Change the key value here.

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,34 @@
...
+
+    bs = beanstalkc.Connection(host="openshift")
+    bs.use("start_lint")
```
navidshaikh: The job needs to be put master tube with action `start_lint` and in `worker_dispatcher.py` read the action and route the job to tube `start_lint` tube, then `worker_dockerfile_lint.py` will read to the tube. This way you are putting the right from this python script to `start_lint` job and `worker_dockerfile_lint.py` is reading it, this way the `worker_dispatcher.py` is bypassed. We should put `worker_dispatcher.py` is stop-relay point for all the jobs and have it as router.

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -57,9 +58,11 @@ def lint_job_data(job_data):
     logger.log(level=logging.INFO, msg="Dockerfile Lint check done")
 
     response = {
-        "logs": out,
+        "logs": None,
+        "linter_results": out,
         "action": "notify_user",
```
navidshaikh: This is going to break the flow, with action `notify_user` and `logs` in the job info, the notify user worker is going to identify the job as failed build and send out the failed build email, with dockerfile linter results as build failure logs in a log file, which is incorrect and job will never get built.

Check the condition about logs and action in this worker https://github.com/CentOS/container-pipeline-service/blob/master/mail_service/worker_notify_user.py#L14

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,34 @@
...
+        filename = filename[1:]
+
+    lint_job_data = {}
+
+    with open(os.path.join(os.environ['DOCKERFILE_DIR'], git_path, filename)) \
+    as f:
+        dockerfile = f.read()
+
+    bs = beanstalkc.Connection(host="openshift")
+    bs.use("start_linter")
```
navidshaikh: We should use `master_tube` here.
The flow should be, every trigger puts the job on `master_tube` with proper action, then `worker_dispatcher.py` reads every job coming over `master_tube` and routes the job based on action to respective tube (`start_linter`). 
So, the rule gonna be, the triggers / invokers put the job on master tube with proper action, `worker_dispatcher.py` acts as stop-relay point for routing the jobs. And all the receivers and readers reads the respective tubes (in your case `start_linter`), and do its job.

In this case, `worker_dispatcher.py` is bypassed, the job is put right from the trigger and received at worker. I dont think its correct.

bamachrn:
LGTM :+1:

kbsingh:
There are still quite a few comments outstanding

dharmit:
@kbsingh are you pointing towards the environment variable thing and resulting `IOError` if file could not be found? Or something else?

mail_service/worker_notify_user.py
```diff
@@ -95,6 +92,58 @@ def notify_user_with_scan_results(job_info):
...
+        print "==> Sending linter results email to user: %s" % notify_email
+        # last parameter (logs) has to None for the sake of
+        # condition put in send_mail function
+        send_mail(notify_email, subject, text, None)
+
+    # We notified user, lets put the job on delivery tube
+    # all other details about job stays same
+    next_job = job_info
+    # change the action
+    next_job["action"] = "start_delivery"
```
navidshaikh: Will this by pass test / scanning phase?

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,54 @@
...
+
+    lint_job_data = {}
+
+    try:
+        dockerfile_location = \
+            os.path.join(os.environ['DOCKERFILE_DIR'], git_path, filename)
+        with open(dockerfile_location) as f:
+            dockerfile = f.read()
+    except IOError:
+        print "==> Error opening the file %s" % dockerfile_location
```
navidshaikh: Log the exact error as well, which you can access as
```
except Exception as e:
   print "=> Error: %s" str(e)
```

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,54 @@
...
+            "notify_email": notify_email,
+            "job_name": job_name,
+            "msg": "Couldn't find the Dockerfile"
+        }
+
+        bs.use("master_tube")
+        jid = bs.put(json.dumps(response))
+        print "==>Put job on 'master_tube' tube with id: %d" % jid
+    else:
+        lint_job_data["name_space"] = name_space
```
navidshaikh: inconsistency in using the key `namespace` for job data.
Earlier we used `namespace`, here we are using `name_space`. This can cause the key error.

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -58,8 +60,12 @@ def lint_job_data(job_data):
```
navidshaikh: Check for `err` here before logging the message "Dockerfile Lint check done" (a line above).

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,54 @@
...
+            os.path.join(os.environ['DOCKERFILE_DIR'], git_path, filename)
+        with open(dockerfile_location) as f:
+            dockerfile = f.read()
+    except IOError:
+        print "==> Error opening the file %s" % dockerfile_location
+        print "==> Sending Dockerfile linter failure email"
+        response = {
+            "linter_results": True,
+            "action": "notify_user",
+            "namespace": name_space,
```
navidshaikh: use consistent variable names; `namespace`

mail_service/worker_notify_user.py
```diff
@@ -95,6 +92,58 @@ def notify_user_with_scan_results(job_info):
...
+
+    # Remove the msg and logs from the job_info as they are not needed now
+    next_job.pop("msg")
+    next_job.pop("logs")
```
navidshaikh: Say linter execution failed, there is no `logs` key in job info then, if you will try to pop a non existent key, it will raise keyerror here.

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,55 @@
...
+        filename = filename[1:]
+
+    lint_job_data = {}
+
+    try:
+        dockerfile_location = \
+            os.path.join(os.environ['DOCKERFILE_DIR'], git_path, filename)
+        with open(dockerfile_location) as f:
+            dockerfile = f.read()
+    except IOError:
```
dharmit: @navidshaikh only reading the files contents would raise an exception (if file doesn't exist). `os.path.join` is more like string concatenation and wouldn't raise any exception AFAIK.

Or is there anything else that you're trying to suggest here?

mail_service/worker_notify_user.py
dharmit: Thanks for pointing. Changing to `start_scan`.

navidshaikh: It should move to `start_build` making sure, all the job parameters for build phase is intact.

mail_service/worker_notify_user.py
dharmit: Right. Adding a check before `pop` to pop only if the key exists.

mail_service/worker_notify_user.py
dharmit: :+1:

client/trigger_dockerfile_lint.py
dharmit: Agreed, adding.

client/trigger_dockerfile_lint.py
dharmit: Agreed, changing.

beanstalk_worker/worker_dockerfile_lint.py
dharmit: Done. Added `if..else` block to handle success & failure situations.

mail_service/worker_notify_user.py
```diff
@@ -95,6 +92,58 @@ def notify_user_with_scan_results(job_info):
...
+    notify_email = job_info['notify_email']
+    project = job_info["namespace"] + "/" + job_info["job_name"]
+
+    subject = "Dockerfile linter results for project: %s" % project
+    if "logs" not in job_info:
+        # Linter failed for some reason which should be in "msg" key
+        text = "Failed to scan the Dockerfile for project %s" % project
+        text += "Reason for failure: %s" % job_info["msg"]
+
+        print "==> Sending linter results email to user: %s" % notify_email
```
dharmit: Done.

mail_service/worker_notify_user.py
navidshaikh: @dharmit : this means, we need to delete the job right after the email is sent.

dharmit:
#dotests

dharmit:
@bamachrn @mohammedzee1000 @navidshaikh @rtnpro need reviews please!

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -28,20 +28,20 @@ def write_dockerfile(dockerfile):
...
 
 
 def lint_job_data(job_data):
     logger.log(level=logging.INFO, msg="Received job data from tube")
     logger.log(level=logging.INFO, msg="Job data: %s" % job_data)
 
-    namespace = job_data.get('name_space')
-
     dockerfile = job_data.get("dockerfile")
+    notify_email = job_data.get("notify_email")
```
navidshaikh: notify_email is not used in the file anywhere ?

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -28,20 +28,20 @@ def write_dockerfile(dockerfile):
...
+    job_name = job_data.get("job_name")
```
navidshaikh: `job_name` also not used in file anywhere?

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -54,24 +54,46 @@ def lint_job_data(job_data):
...
-        msg="Put job on 'master_tube' tube with id: %d" % jid
-    )
+    if err is None:
+        logger.log(level=logging.INFO, msg="Dockerfile Lint check done")
+        response = {
+            "logs": out,
+            "linter_results": True,
+            "action": "notify_user",
+            "namespace": job_data.get('namespace'),
+            "notify_email": notify_email,
```
navidshaikh: if notify_email is not used anywhere, you can use, `job_data.get("notify_email")` here

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -54,24 +54,46 @@ def lint_job_data(job_data):
...
+            "job_name": job_name,
```
navidshaikh: (same as above)

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -54,24 +54,46 @@ def lint_job_data(job_data):
...
+            level=logging.INFO,
+            msg="Put job on 'master_tube' tube with id: %d" % jid
+        )
+    else:
+        logger.log(level=logging.ERROR, msg="Dockerfile Lint check failed")
+        response = {
+            "linter_results": True,
+            "action": "notify_user",
+            "namespace": job_data.get('namespace'),
+            "notify_email": notify_email,
```
navidshaikh: pass the `notify_email` value right from the job_data. If we are not using it in the code, we don't need to create a var and pass again.

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -54,24 +54,46 @@ def lint_job_data(job_data):
...
+            "job_name": job_name,
+            "msg": err
+        }
+
+        bs.use("master_tube")
+        jid = bs.put(json.dumps(response))
+        logger.log(
+            level=logging.INFO,
+            msg="Put job on 'master_tube' tube with id: %d" % jid
+        )
```
navidshaikh: Irrespective of the result of linter, we are putting the job on master tube, this block of code can be taken out of `if` `else`

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,55 @@
...
+        print "==> Error opening the file %s" % dockerfile_location
+        print "==> Error: %s" % str(e)
+        print "==> Sending Dockerfile linter failure email"
+        response = {
+            "linter_results": True,
+            "action": "notify_user",
+            "namespace": namespace,
+            "notify_email": notify_email,
+            "job_name": job_name,
+            "msg": "Couldn't find the Dockerfile at specified git_path"
```
navidshaikh: we can also put the path `dockerfile_location` in `msg`, helps debugging better.

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,55 @@
...
+        bs.use("master_tube")
+        jid = bs.put(json.dumps(response))
+        print "==>Put job on 'master_tube' tube with id: %d" % jid
+    else:
+        lint_job_data["namespace"] = namespace
+        lint_job_data["job_name"] = job_name
+        lint_job_data["dockerfile"] = dockerfile
+        lint_job_data["notify_email"] = notify_email
+        lint_job_data["action"] = "start_linter"
+
```
navidshaikh: did we forget here to use the master tube?

```
bs.use("master_tube")
bs.put(json.dumps(lint_job_data))
```

client/trigger_dockerfile_lint.py
```diff
@@ -0,0 +1,55 @@
...
+        bs.put(json.dumps((lint_job_data)))
```
navidshaikh: extra `()`

client/trigger_dockerfile_lint.py
dharmit: Tried that but somehow it's not appearing in the resulting emails sent out to the user.

client/trigger_dockerfile_lint.py
dharmit: Nope. `bs.use("master_tube")` is already there at the top of the file. It's used unnecessarily in the `except` block. Removed it.

client/trigger_dockerfile_lint.py
dharmit: In any case, the value doesn't show up in email.

navidshaikh: I think the control is going in this block, only because `dockerfile_location` value is not found. We need to find out why!

PR #91, opened 2016-10-13
Fixes invocation of scanner and bug in mail service worker

navidshaikh:
Fixes invocation of atomic scanner in test phase
  Fixes creating mount directory and adds error handling
  Fixes error in mail service worker

### DIFF
--- a/beanstalk_worker/worker_start_scan.py
+++ b/beanstalk_worker/worker_start_scan.py
@@ -5,6 +5,7 @@ import docker
 import json
 import logging
 import os
+import shutil
 import subprocess
 import sys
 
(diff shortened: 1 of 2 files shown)


bamachrn:
#dotests

beanstalk_worker/worker_start_scan.py
```diff
@@ -96,18 +97,43 @@ def scan_job_data(job_data):
...
+        mount_object.unmount()
+        shutil.rmtree(image_rootfs_path)
+        try:
+            os.makedirs(image_rootfs_path)
+        except OSError as error:
+            logger.log(
+                level=logging.CRITICAL,
+                msg=str(error)
+                )
+            raise
```
bamachrn: @navidshaikh should we raise the error or we should handle it to write in logs? as if it raises the exception without handling, it will cause the worker to stop working for next jobs also.

navidshaikh: Sure! Updated the PR.

bamachrn:
LGTM

PR #99, opened 2016-10-21
Patch atomic scanner to run on non tty input.

rtnpro:
We are applying this patch because atomic cli v1.11 is not yet available in CentOS repo.

Patching from https://github.com/projectatomic/atomic/commit/a261270e727353899675a3fde8f9108376bb27cf

### DIFF
--- a/provisions/roles/scanner/tasks/main.yml
+++ b/provisions/roles/scanner/tasks/main.yml
@@ -3,6 +3,14 @@
   yum: name=atomic state=present
   sudo: yes
 
+# Patch from https://github.com/projectatomic/atomic/commit/a261270e727353899675a3fde8f9108376bb27cf
+- name: Patch atomic scanner to run on non tty input
+  replace: >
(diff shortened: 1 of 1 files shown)


rtnpro:
I need to test it out, once again, please don't merge until then.

provisions/roles/scanner/tasks/main.yml
```diff
@@ -3,6 +3,13 @@
   yum: name=atomic state=present
   sudo: yes
 
+# Patch from https://github.com/projectatomic/atomic/commit/a261270e727353899675a3fde8f9108376bb27cf
+- Patch atomic scanner to run on non tty input
```
navidshaikh: name: Patch atomic scanner to run on non tty input

rtnpro:
@navidshaikh thanks for the review. I have updated the code based on your feedback.

mohammedzee1000:
LGTM

PR #104, opened 2016-11-11
Updated pattern for ADD REGISTRY regex

bamachrn:
Adding registry and insecure registry in /etc/sysconfig/docker was not getting replaced due to wrong pattern match.  fixed the same.


dharmit:
LGTM

PR #107, opened 2016-11-16
Updates info about pipeline-scanner

navidshaikh:


navidshaikh:
@dharmit : ping

atomic_scanners/pipeline-scanner/pipeline-scanner
```diff
@@ -3,8 +3,8 @@ scanner_name: pipeline-scanner
 image_name: registry.centos.org/pipeline-images/pipeline-scanner
 default_scan: redhat-release
```
dharmit: This needs to be changed to the value of `name`.

atomic_scanners/pipeline-scanner/pipeline-scanner
```diff
@@ -3,8 +3,8 @@ scanner_name: pipeline-scanner
...
 scans: [
-  { name: redhat-release,
+  { name: pipeline-scanner,
```
dharmit: I'd suggest to use a name like `yum-update` or `rpm-update` since that's what this scan does. Using the name `pipeline-scanner` wouldn't indicate to the user as to what it will do.

navidshaikh: We have been using pipeline-scanner in image name as well, I think description is providing info about the scanner, IMO, we should keep the name intact and let the description detail about its scope.
OR
We might want to change the name to `yum-update` throughout for image name, dir name for scanner and all other places.

dharmit: Image name `pipeline-scanner` will not be affected by changing the name of a scan. In fact, I think, it won't be affected even if we changed the scanner name. Because we're specifying the image to be used.

Adding detail in the description sounds ok to me but if we think about it, scan named `pipeline-scanner` would only provide yum update facility. Doesn't sound appropriate to me.

Same image can be used to do more than one type of scan under the scanner `pipeline-scanner`. As a result, having appropriate name for a scan type makes more sense to me. Here's an example of single image doing multiple scans:

~~~
$ atomic scan --list
Scanner: openscap
  Image Name: rhel7/openscap
     Scan type: cve *
     Description: Performs a CVE scan based on known CVE data

     Scan type: standards_compliance
     Description: Performs a standard scan


* denotes defaults
~~~

navidshaikh: https://github.com/CentOS/container-pipeline-service/blob/master/beanstalk_worker/worker_start_scan.py#L346 , if we change the name, worker wont be able to find the scanner.

bamachrn: I agree with @dharmit we should have separate names for all the scanners like rpm-verify, yum-update or rpm-update. This will help us differentiating scanners. All the scanners fall under pipeline-scanner, so may be we can change to make it more specific

navidshaikh:
@dharmit: kindly take a look.

atomic_scanners/pipeline-scanner/pipeline-scanner
```diff
@@ -1,10 +1,10 @@
 type: scanner
 scanner_name: pipeline-scanner
 image_name: registry.centos.org/pipeline-images/pipeline-scanner
-default_scan: redhat-release
+default_scan: pipeline-scanner
 scans: [
-  { name: redhat-release,
+  { name: pipeline-scanner,
```
dharmit: What I'm suggesting is to have a name like `yum-update` or `rpm-update`. Reason being that additional scan types we plan to include for the pipeline will come under the same scanner, that is, `pipeline-scanner`.  Naming the scan type to `pipeline-scanner` doesn't help one understand the purpose of the scan.

I understand that the way [scanner worker is currently designed](https://github.com/CentOS/container-pipeline-service/blob/master/beanstalk_worker/worker_start_scan.py#L346), this would break stuff. We need to add scan types to the command we're executing to get proper scan results. This can be done by passing `--scan-type=<name>` to the command. That should help take care of things when, in future, we add more scan types.

Also, at the moment, this won't break the worker because we've got only one scan type under the `pipeline-scanner` and it's default. So, `atomic scan` will execute that without us providing any scan type to it.

atomic_scanners/pipeline-scanner/pipeline-scanner
```diff
@@ -1,10 +1,10 @@
...
     args: ['python', 'scanner.py', 'release'],
-    description: "Find Red Hat release"
```
navidshaikh: @dharmit : that will require changing the scanner itself and worker a bit, but meanwhile we want to keep the wrong description for this scanner?

dharmit: Right that will require us to change the scanner itself as well along with the worker. But we can take care of it when we start adding additional scan type to it. In the meanwhile, changing the `name` of scan is not going to involve any additional work.

Umm, I didn't mention anything about the `description`. That looks good to me! :+1:

dharmit:
LGTM

navidshaikh:
Verified the PR by building scanner image and spinning it, rebased to latest master, good for merge. @dharmit  ping

PR #115, opened 2016-12-01
Trigger weekly scan

### LINKED ISSUE #109: Trigger weekly scan of images in the registry
All the images in the registry should be scanned on weekly basis. Results of the scan should be sent to the concerned developer via email.

dharmit:
This PR fixes #109. A weekly scan of all the images in the registry would be triggered at 23:59:00 on Saturday.

### DIFF
--- a/jenkinsbuilder/weekly-scan.py
+++ b/jenkinsbuilder/weekly-scan.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python
+
+import beanstalkc
+import glob
+import json
+import os
+import subprocess
+import sys
+import yaml
+
+
+# connect to beanstalkd tube
+bs = beanstalkc.Connection("BEANSTALK_SERVER")
+bs.use("master_tube")
+
+# registry server value to be replaced by ansible
+registry = "JENKINS_SLAVE"
+
+# cli argument pointing to directory containing yml files
+index_dir = sys.argv[1]
+
+# string representation of the catalog on r.c.o
+str_catalog = subprocess.check_output(
+    ["curl", "%s:5000/v2/_catalog" % registry]
+)
+
+# convert string into a list to use for comparison
+json_catalog = json.loads(str_catalog).values()[0]
+
+# have a list of files in the index_dir
+files = glob.glob("%s/*.yml" % index_dir)
+
+# index dir will always have yml files only; but just in case
+for f in files:
+    if f.endswith(".yml"):
+        continue
+    files.remove(f)
+
+# parse the yml file
+for f in files:
+    with open(os.path.join(os.environ.get("CWD"), "index.d", f)) as stream:
+        try:
+            yaml_parse = yaml.load(stream)
+        except yaml.YAMLError as e:
+            print e
+
+    # all entries in a yml file in list of dictionaries format
+    entries = yaml_parse.values()[0]
+
+    for entry in entries:
+        app_id = entry["app-id"]
+        job_id = entry["job-id"]
+        desired_tag = entry["desired-tag"]
+        email = entry["notify-email"]
+
+        entry_short_name = app_id + "/" + job_id
+
+        # Scan an image only if it exists in the catalog!
+        if entry_short_name in json_catalog:
+            data = {
+                "action": "start_scan",
+                "tag": desired_tag,
+                "name_space": app_id + "-" + job_id + "-" + desired_tag,
+                "name": "%s:5000/%s/%s:%s" %
+                (registry, app_id, job_id, desired_tag),
+                "notify_email": email,
+                "weekly": True
+            }
+
+            job = bs.put(json.dumps(data))
+            print "Image %s sent for weekly scan with data %s" % \
+                (entry_short_name, data)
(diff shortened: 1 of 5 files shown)


dharmit:
@bamachrn not sure why it's showing so many commits! The only one that matters for this PR is [cc41d3b](https://github.com/CentOS/container-pipeline-service/pull/115/commits/cc41d3b1c483a28e30a7a1100b083d1b69761048). Trying a rebase says that branch is already up to date.

beanstalk_worker/worker_start_scan.py
```diff
@@ -110,7 +110,8 @@ def run(self):
 
         # TODO: Figure out why random tag (with date) is coming
         # image_under_test = ":".join(self.job_info.get("name").split(":")[:-1])
-        image_under_test = self.job_info.get("name")
+        image_under_test = self.job_info.get("name") + ":" \
+            + self.job_info.get("tag")
```
dharmit: @bamachrn @navidshaikh need your comments on this line. I'm worried it'd do some harm and not always work as expected! Because there are instances where we specify tag info in the name key itself.

navidshaikh: @dharmit : the same code will be executed for usual scan per build and weekly scan as well, I'd suggest you test this PR along with triggering usual scans per build as well and verify.

jenkinsbuilder/weekly-scan.py
```diff
@@ -0,0 +1,71 @@
...
+str_catalog = subprocess.check_output(
+    ["curl", "%s:5000/v2/_catalog" % registry]
+)
+
+# convert string into a list to use for comparison
+json_catalog = json.loads(str_catalog).values()[0]
+
+# have a list of files in the index_dir
+for directory, dirnames, filenames in os.walk(index_dir):
+    files = filenames
```
navidshaikh: if the loop to iterate for once, we can use other methods as well instead of walk

jenkinsbuilder/weekly-scan.py
```diff
@@ -0,0 +1,71 @@
...
+
+# index dir will always have yml files only; but just in case
+for f in files:
+    if f.endswith(".yml"):
+        continue
+    files.remove(f)
```
navidshaikh: okay!

jenkinsbuilder/weekly-scan.py
```diff
@@ -0,0 +1,71 @@
...
+
+# parse the yml file
+for f in files:
+    with open(os.path.join(os.environ.get("CWD"), "index.d", f)) as stream:
+        try:
+            yaml_parse = yaml.load(stream)
+        except yaml.YAMLError as e:
+            print e
```
navidshaikh: and log it as well somewhere, we'll not have record otherwise if something failed.

jenkinsbuilder/weekly-scan.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # Scan an image only if it exists in the catalog!
+        if entry_short_name in json_catalog:
+            data = {
+                "action": "start_scan",
+                "tag": desired_tag,
+                "name_space": app_id + "-" + job_id + "-" + desired_tag,
+                "name": registry + ":5000/" + app_id + "/" + job_id,
+                "notify_email": email,
+                "weekly": True
```
navidshaikh: nice! we are using the tube per our requirements and use case.

beanstalk_worker/worker_start_scan.py
dharmit: Yep, I did that just before making earlier comment. And I'm getting proper mail notifications for at least one image that I built which doesn't use `latest` tag.

jenkinsbuilder/weekly-scan.py
dharmit: Hmm, `os.listdir()` seems to be a better choice to me. You have any other suggestions?

jenkinsbuilder/weekly-scan.py
dharmit: I think we should log it to some file when the NFS is up. Because this doesn't run as a service. It runs as a one time script once in a week. WDYT?

With the `print` statement, we'll have record in the Jenkins console output for a build.

dharmit:
> After the image is sent for weekly scan with action scan, the current behaviour is, we relay the job to delivery phase after scan, since the image is just built, however in case of weekly scan, the image was already built and we do not need to push the job to delivery phase after the scan, it is unnecessary load on our engines. I suggest, we make scan worker aware of weekly scan, and do not relay the job for delivery phase and delete the job right after scan and notification.

Because of this, system was unnecessarily triggering success emails. So, triggering a weekly scan would send out scan results as well as success building image emails. Latter being out of error.

Corrected the same.

mail_service/worker_notify_user.py
```diff
@@ -75,6 +78,11 @@ def notify_user_with_scan_results(job_info):
     # condition put in send_mail function
     send_mail(notify_email, subject, text, None)
 
+    # if weekly scan is being executed, we do not want trigger delivery phase
+    if job_info.get("weekly"):
+        print "Weekly scan completed; moving to next job."
+        return
+
```
navidshaikh: nice!

navidshaikh:
Hold on for merge until @dharmit  ACKs after integration testing.

bamachrn:
LGTM 👍

PR #117, opened 2016-12-06
Added LetsEncrypt support

arrfab:
Added some proxy_pass to allow external/central node to be used as TLS letsencrypt client node (and not the nginx node)


bamachrn:
#dotests

PR #119, opened 2016-12-08
Added upstream.registry.centos.org for vhost

arrfab:
Added upstream.registry.centos.org as hostname for vhost (that will permit the letsencrypt cert to be generated at the same time as for the newer registry.centos.org one)


bamachrn:
#dotests

bamachrn:
LGTM

PR #120, opened 2016-12-14
Filters documentation files from RPM verify output

navidshaikh:
Like configuration files, the documentation files are also now filtered.
 With documentation files in image, the image size increases, by removing
 the documentation files from RPM installed files, image size decreases.


dharmit:
LGTM

PR #121, opened 2016-12-20
Fix weekly scan string concatenation issue

dharmit:
This fixes the bug that prevented weekly scan from running in prod for the first time.


navidshaikh:
> This fixes the bug that prevented weekly scan from running in prod for the first time.

What was the issue? unicode chars ?

dharmit:
> What was the issue? unicode chars ?

String concatenation. `str` and `int` values were being concatenated and failed. So made sure that all possible places where there could be an integer value, get converted to `str` first.

dharmit:
#dotests

PR #123, opened 2016-12-21
Moved basic setup to script and cleaned up the code.

mohammedzee1000:
Fix issue with index ci failing if desired-tag is a number and reenable index ci test runs which are currently failing.

Also, did cleanup of the code with a lot of refactoring to make the code more manageable.


bamachrn:
#dotests

mohammedzee1000:
@bamachrn @dharmit @navidshaikh  comments, reviews please

bamachrn:
LGTM, but this would be a temporary fix

bamachrn:
#dotests

rtnpro:
#dotests

mohammedzee1000:
Once the job.yml thing eats up job.yml correctly, we can remove the 2 lines being added to init

testindex/setup_basic.py
```diff
@@ -0,0 +1,30 @@
...
+import subprocess
+
+the_os = platform.dist()[0]
+
+
+def run_cmd(cmd):
+    """Runs a command and waits for it to finish"""
+    process = subprocess.Popen(cmd, shell=True)
+    process.wait()
+    return process.returncode
```
dharmit: We're returning the exit code of the command but not using it anywhere.

mohammedzee1000: I could not think of a use when i wrote this, but considered the possibility of its use later.
I will see if i can include the check for it in the script

mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

bamachrn:
#dotests

testindex/__init__.py
```diff
@@ -4,21 +4,14 @@
...
     parser.add_argument("-i",
-                        "--indexd_location",
-                        help="Specify location of indexd. Default is ./index.d",
-                        metavar="INDEXD",
-                        nargs=1,
-                        action="store")
+                        "--index",
+                        help="Specify location of index. Default is ./index.d",
+                        metavar="INDEX_PATH",
+                        default="./index.d")
```
dharmit: For the sake of readability (because that's quite important in Python), can we have something like:

```python
parser.add_argument(
    "-i",
    "--index",
    help="Specify location of index. Default is ./index.d",
    metavar="INDEX_PATH",
    default="./index.d"
)
```

testindex/__init__.py
```diff
@@ -28,7 +21,7 @@ def get_parser():
     parser.add_argument("-c",
                         "--cleanup",
                         help="Specify to force cleanup of test bench. If your running multiple times on same system,"
-                             " this is not recommended as it will require downloading of repos again and again",
+                             " this is not recommended as it will require downloading of repositories again and again",
                         action="store_true")
```
dharmit: Same as readability comment above.

testindex/summary.py
```diff
@@ -21,39 +21,46 @@ def add_warning(self, msg):
...
     """This class summarizes the tests"""
 
-    _summary = {}
-
-    def __init__(self):
+    def __init__(self, summary_dump):
+        self.global_errors = []
+        self._summary = {}
+        self._summary_dump = summary_dump
         pass
```
dharmit: Forgot to remove the `pass` statement here?

testindex/summary.py
```diff
@@ -21,39 +21,46 @@ def add_warning(self, msg):
...
         entry_hash = Summary._get_entry_hash(file_name, entry)
 
-        if entry_hash not in Summary._summary[file_name]:
-            Summary._summary[file_name][entry_hash] = {
+        if entry_hash not in self._summary[file_name]:
+            self._summary[file_name][entry_hash] = {
                 "entry": str(entry),
                 "errors": [],
                 "warnings": []
             }
```
dharmit: That's the readability I mentioned in few other comments. Thanks for this! :+1:

testindex/summary.py
```diff
@@ -21,39 +21,46 @@ def add_warning(self, msg):
...
         print "\n####################### SUMMARY ##################\n"
 
-        for file_name, entries in Summary._summary.iteritems():
+        print "\nGLOBAL ERRORS:\n"
+        if len(self.global_errors) == 0:
+            print "NONE\n"
+        else:
+            for err in self.global_errors:
+                print "**E " + err
+            print
```
dharmit: Maybe just do:
```python
print "**E " + err + "\n"
```
instead of a blank print that would also just insert a newline? 

Not a big deal if we don't. :smile:

testindex/nuts_and_bolts.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if not path.exists(self.generator_dir):
+            mkdir(self.generator_dir)
+
+        if path.exists(self.generator_dir):
+            self._cleanup_content(self.generator_dir)
+
+        if not path.exists(self.generator_dir):
+            mkdir(self.generator_dir)
+
```
mohammedzee1000: @dharmit thats why i have multiple checks and mkdirs, i am first checking parent dir and if its not present, making it and then proceeding to next child dir and so on

testindex/Engine.py
```diff
@@ -1,48 +1,51 @@
...
-import Validators
-from NutsAndBolts import Environment, GlobalEnvironment, StatusIterator
-from Summary import Summary
+import validators
+from nuts_and_bolts import Context, StatusIterator
 
 
 class Engine:
-    def __init__(self, index_location="./index.d", dump_location="cccp-index-test", cleanup=False):
+    def __init__(self, index_path="./index_path.d", cleanup=False, silent=False):
```
mohammedzee1000: I will into how i can get that in :)

testindex/nuts_and_bolts.py
dharmit: I'm assuming you're talking about the last point of having `for` loop here. A list's order won't be changed. So you can use the same order when looping over a list as you're using now by individual `if..else` statements.

Would like to have your thoughts on exception handling as well.

mohammedzee1000: Ok i see your point though i dont need 2 checks for generator dir anymore since i am now time stamping it and hence each run gets its own generator dir. So i am surrounding it with a for loop and try catch.

testindex/summary.py
mohammedzee1000: That is to introduce a new line at and end of all the errs being printed.

testindex/summary.py
mohammedzee1000: I strive for readability infact i will be doing a lot more refactoring for making the code more readable in later prs. :)

dharmit:
LGTM, thanks for making requested changes so fast!

bamachrn:
LGTM, merging this

PR #126, opened 2016-12-21
Reduced layers for builder images

bamachrn:
* Reducing extra layers for builder images for faster provisioning


dharmit:
LGTM, will merge post CI check.

PR #130, opened 2016-12-22
Add atomic scanner for pip, npm and gem package managers

dharmit:


dharmit:
#dotests

dharmit:
In this PR, I've added a new `beanstalk_worker/scanner.py` which defines a class `Scanner`. It acts as a base class and can be used by the classes for individual scanners inside `beanstalk_worker/worker_start_scan.py`. An example of this is the `MiscPackageUpdates` class which uses the methods of base class to perform atomic scan.

@navidshaikh We can modify `ScannerRPMVerify` class to use the same as base class and remove duplicate code.

@bamachrn @mohammedzee1000 @navidshaikh need PR review.

beanstalk_worker/scanner.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
```
bamachrn: this file is the central scanner right?

atomic_scanners/misc-package-updates/README.md
```diff
@@ -0,0 +1,30 @@
+Atomic scanner: misc-package-updates
+--------------------------------
+
+This is a container image scanner based on `atomic scan`. The goal of the
+scanner is to scan CentOS based Docker images in the CentOS Community Container
```
navidshaikh: *CentOS based container images

atomic_scanners/misc-package-updates/README.md
```diff
@@ -0,0 +1,30 @@
...
+Pipeline and generate relevant results.
+
+Steps to use:
+
+- Pull Docker image from **registry.centos.org**:
```
navidshaikh: *Pull container image from **registry.centos.org**

atomic_scanners/misc-package-updates/scanner.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def remove_last_blank_character(response):
+    # last value of response generated above could be ''
+    if response[-1] == '':
+        response = response[:-1]
+    return response
+
+
+def list_of_outdated_packages(response):
```
navidshaikh: docstrings

atomic_scanners/misc-package-updates/scanner.py
```diff
@@ -0,0 +1,164 @@
...
+            "Could not find {} executable in the image".format(cli_arg)
+    else:
+        json_out["Scan Results"]["{} package updates".format(cli_arg)] = \
+            format_response(cli_arg, response)
+        json_out["Finished Time"] = \
+            datetime.now().strftime('%Y-%m-%d-%H-%M-%S-%f')
+
+    # remove the container
+    client.remove_container(container=container.get("Id"), force=True)
+except Exception as e:
```
navidshaikh: log the exception /error message and we should put another error message about failure.

provisions/roles/scanner/tasks/main.yml
```diff
@@ -27,6 +27,10 @@
     regexp="^#*\s*ADD_REGISTRY=.*"
     replace='ADD_REGISTRY="--add-registry {{ public_registry }} --insecure-registry {{ public_registry }}:5000"'
 
+- name: Set SELinux to permissive
+  selinux:
+      state: permissive
+
```
navidshaikh: why is this needed?

atomic_scanners/misc-package-updates/misc-package-updates
```diff
@@ -0,0 +1,20 @@
+type: scanner
+scanner_name: misc-package-updates
+image_name: registry.centos.org/pipeline-images/misc-package-updates
```
dharmit: @bamachrn yes, it'll be built with latest tag.

beanstalk_worker/scanner.py
dharmit: @bamachrn yep, it's the base class that would be inherited by other scanners.

provisions/roles/scanner/tasks/main.yml
dharmit: @navidshaikh because we're mounting Docker socket `/var/run/docker.sock` inside the container that's going to scan the image. Using which a new container is spunn off to check for respective package manager updates.

atomic_scanners/misc-package-updates/scanner.py
dharmit: Updated. Not sure why GitHub doesn't say "Show outdated".

atomic_scanners/misc-package-updates/scanner.py
dharmit: Updated. Not sure why GitHub doesn't say "Show outdated".

atomic_scanners/misc-package-updates/scanner.py
```diff
@@ -0,0 +1,208 @@
...
+            cmd="gem outdated"
+        )
+
+        response = client.exec_start(exe)
+
+    # Check for npm updates
+    elif cli_arg == "npm":
+        exe = client.exec_create(
+            container=container.get("Id"),
+            cmd="npm outdated"
```
msrb: I am not sure if this command can do the trick. There can be modules installed globally (`npm list -g`) and also in `npm_modules` directories located basically anywhere on the filesystem.

dharmit: @msrb agreed. This one takes care of globally installed `npm` modules. However, I doubt we can go through all the directories on the filesystem.

Also, what you mention about `npm` holds true for `pip` as well. If a dev prefers to use `virtualenv` inside the container, we are going to face similar issue.

It boils down to whether a dev wants to use `npm`/`pip` inside container in such a way that they can have multiple environments or a global environment. Our scanner should handle the latter scenario. Not sure if we can handle the former effectively.

msrb: @dharmit I am no NPM expert, so don't quote me on this :) I *think* if you want to check globally installed modules, you need to add `-g`, i.e.: `npm outdated -g`. Otherwise you're checking modules installed in `./npm_modules/` directory.

dharmit: @msrb gotcha. I'm not an `npm` expert either. @bamachrn any thoughts?

dharmit: @msrb I hate to admit it but I hadn't looked at the `npm_modules` dir that gets created whenever you do a `npm install <package>@<version>`. The `-g` tip is more than helpful. It's savior because that's what I actually wanted to accomplish. Having worked with `pip`, I didn't realize `npm` would create a dir at whatever location you execute `npm install` if it doesn't have a `-g` switch.

Thank you!

navidshaikh:
#dotests

dharmit:
#dotests

PR #131, opened 2016-12-26
Fix ci failure due to environment variables not being set properly

mohammedzee1000:


bamachrn:
#dotests

bamachrn:
#dotests

bamachrn:
#dotests

mohammedzee1000:
I dont think this PR can be validated through ci as its change is in the job.yml. So the PR will need to be merged before we can validate it through ci itself.

In the future, we might want to figure out a way to auto validate changes to job.yml.

bamachrn:
PR is run after a merge so it should ideally work

bamachrn:
#dotests

PR #133, opened 2016-12-28
Fix ci builds

rtnpro:
This implements:

- build failures in CICO due to missing PYTHONPATH
- move inline build scripts from  JJB file to external files


bamachrn:
LGTM

Issue #44, opened 2016-08-28
Jenkins jobs for projects not able to trigger build in Openshift in CI


rtnpro:
The  build script for ``bamachrn-python`` project generated in CI looks like:

```bash
cd /opt/cccp-service/client
./build-project.sh bamachrn python https://github.com/bamachrn/cccp-python master / Dockerfile.test bamachrn@gmail.com None
```

The output of the above command is:
```console
[root@n16 client]# ./build-project.sh bamachrn python https://github.com/bamachrn/cccp-python master / Dockerfile.test bamachrn@gmail.com None
build-project.sh NAME REPO_URL
   NAME      Name of the project/namespace
   TAG       Name of the resulting image (image will be named NAME/TAG:latest)
   REPO_URL  URL of project repository containing Dockerfile
   REPO_BRANCH Branch of the repo to be built
   REPO_BUILD_PATH  Relative path to the Dockerfile in the repository
   TARGET_FILE  Name for the dockerfile to be built
   NOTIFY_EMAIL  Email ID to be notified after successful build
   DEPENDS_ON  Dependency list for the current image
   DESIRED_TAG  Tag for the final output image like latest
```

It seems like there are less arguments supplied to the build script. On a more detailed observation, I found that currently, the build script accepts 9 args, where as in 
https://github.com/CentOS/container-pipeline-service/blob/master/jenkinsbuilder/project-defaults.yml#L16
it generates only 8 args.

bamachrn:
missed to push the changes to jenkinsbuilder/projects_default.yml

PR #137, opened 2016-12-30
Added log retrieval in try except

bamachrn:
* this is for for making sure the loop continues even if the log is not available

### DIFF
--- a/beanstalk_worker/worker_start_build.py
+++ b/beanstalk_worker/worker_start_build.py
@@ -112,13 +112,19 @@ def start_build(job_details):
         # checking logs for the build phase
         log_command = "oc logs --namespace " + namespace + " build/" + \
             build_details + kubeconfig
-        logs = run_command(log_command)
-        logs = logs[0]
+        try:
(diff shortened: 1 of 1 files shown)


beanstalk_worker/worker_start_build.py
```diff
@@ -112,13 +112,19 @@ def start_build(job_details):
         # checking logs for the build phase
         log_command = "oc logs --namespace " + namespace + " build/" + \
             build_details + kubeconfig
-        logs = run_command(log_command)
-        logs = logs[0]
+        try:
+            logs = run_command(log_command)
+            logs = logs[0]
```
dharmit: If line #44:

```python
out = p.communicate()
```

is changed to:

```python
out, err = p.communicate()
```

we'll not need to do:
```python
logs = logs[0]
```

dharmit:
LGTM, merging.

bamachrn:
@dharmit rebased and done changes

dharmit:
LGTM, merging.

PR #138, opened 2017-01-03
Fixed fetching exit code from remote cmd run streaming output

rtnpro:


rtnpro:
If the build fails, then this PR is ready to merge.

PR #139, opened 2017-01-03
Moved logs to failed builds only

bamachrn:
* Log command was blocking the builds if the size of the logs is large.
* for now we are only sending logs in failed builds 
once we take NFS in place these problems will be resolved


rtnpro:
LGTM :+1: but we need to wait for the build result.

PR #125, opened 2016-12-21
Logs UI

navidshaikh:
Fixes #124 
Fixes #141

PR #144, opened 2017-01-04
Delete job if scanner data is too big

dharmit:
This is a temporary fix till the time we modify the workers to use NFS volumes to store data instead of sending it over the beanstalkd tube.

Need PR review; ping @rtnpro @bamachrn @mohammedzee1000 @navidshaikh

### DIFF
--- a/beanstalk_worker/worker_start_scan.py
+++ b/beanstalk_worker/worker_start_scan.py
@@ -433,5 +433,11 @@ while True:
             msg="Job moved from scan phase, id: %s" % job_id
             )
         job.delete()
+    except beanstalkc.CommandFailed as e:
+        logger.log(
+            level=logging.ERROR,
+            msg="Failed to put scanner data on beanstalkd tube; deleting job"
+        )
+        job.delete()
     except Exception as e:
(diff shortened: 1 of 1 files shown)


navidshaikh:
#dotests

beanstalk_worker/worker_start_scan.py
```diff
@@ -433,5 +433,11 @@ def process_output(self, json_data):
             msg="Job moved from scan phase, id: %s" % job_id
             )
         job.delete()
+    except beanstalkc.CommandFailed as e:
+        logger.log(
+            level=logging.ERROR,
+            msg="Failed to put scanner data on beanstalkd tube; deleting job"
+        )
+        job.delete()
```
navidshaikh: deleting the job is needed in either case, whether scanner ran successfully or there was an exception, deleting the job in `finally` clause would make more sense. With logs UI feature, this is implemented in the https://github.com/CentOS/container-pipeline-service/pull/125 with additional code.

navidshaikh:
We have NFS PR merged in master, this was created as temporary fix.

PR #146, opened 2017-01-11
WIP - Update linter image before execution

dharmit:
In this PR, we're modifying the Dockerfile lint worker to pull in the linter image right before linting the Dockerfile. This should ensure we have a latest image every time.

### DIFF
--- a/beanstalk_worker/worker_dockerfile_lint.py
+++ b/beanstalk_worker/worker_dockerfile_lint.py
@@ -7,6 +7,8 @@ import os
 import subprocess
 import sys
 
+from constants import LINTER_IMAGE
+
 logger = logging.getLogger("cccp-linter")
 logger.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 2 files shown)


dharmit:
@bamachrn @navidshaikh @mohammedzee1000 PR review please? :smiley:

bamachrn:
LGTM 👍

beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -32,6 +34,34 @@ def write_dockerfile(dockerfile):
...
+    out, err = subprocess.Popen(
+        [
+            "docker",
+            "pull",
+            LINTER_IMAGE
+        ],
+        stdout=subprocess.PIPE
+    ).communicate()
+
+    if err is not None:
```
navidshaikh: `if err:`

navidshaikh:
IMO, updating the tooling images should be taken care by piece managing the deployment. This will add some time delay in job execution. It's good to run the updated linter tooling image, however it can increase the total build time. We should think about tradeoff, if the linter image is not really going to be updated very often, hand over the responsibility to some dedicated ansible module.

dharmit:
@navidshaikh I agree. Linter image's not gonna be updated too often. Merging this PR would add time delay of couple of seconds before the linting process begins (that in case of no update to the image in registry.)

dharmit:
Need to work on Ansible provisioning to have a cron job instead of a function that pulls in new container image.

dharmit:
I'm closing this PR in favor of a new one with Ansible changes. It'll handle updating images for all workers that need to use latest images.

PR #148, opened 2017-01-17
Fixed issue in regex pattern match for depends on.

mohammedzee1000:
Dots included after : were taken as mismatch and this has been
fixed.


mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

PR #149, opened 2017-01-17
Cron job to pull fresh container images on daily basis

dharmit:


dharmit:
@rtnpro @bamachrn can you review and help merge this?

bamachrn:
LGTM lets get the logging merged first

dharmit:
#dotests

dharmit:
#dotests

kbsingh:
-1 to the code, 

these sort of things should be trigger based, and setup to update when new containers are deployed rather than from cron.

dharmit:
We explored using the [Docker Hub/Registry Notification plugin](https://github.com/jenkinsci/dockerhub-notification-plugin) for event driven approach. However, the plugin is not implemented for use with JJB which we use to create jobs on Jenkins.

I've created a [story](https://storyboard.openstack.org/#!/story/2000988) on OpenStack storyboard but it depends on this getting priority.

So, now, going with the approach of force pulling a new image during every deployment since deployments happen more frequently than update happens in any of these images.

PR #154, opened 2017-01-25
[wip] Drop epel release

kbsingh:
you can use a process like this, to ship our own epel.repo file, as well as the GPG key for it. This would work around the MM flakyness often seen breaking the ci jobs and yum operations in prod.


mohammedzee1000:
If i want to make use of this in my indexci what would i need to do, i need the python networkx package

jperrin:
If ansible 2.x is a requirements, you can provide a bit more configuration magic via the ansible repo definitions -> https://docs.ansible.com/ansible/yum_repository_module.html

PR #155, opened 2017-01-30
Remove dependency on EPEL repo on CI

rtnpro:
- use centos-admin repo instead
- install python packages for jenkins-job-builder using pip inside virtualenv.


rtnpro:
#dotests

provisions/group_vars/all
```diff
@@ -7,3 +7,5 @@ cccp_source_repo: https://github.com/centos/container-pipeline-service
 cccp_source_branch: master
 
 public_registry: registry.centos.org
+
+enable_epel: true
```
navidshaikh: should this be true ?

rtnpro:
#dotests

rtnpro:
#dotests

provisions/group_vars/all
mohammedzee1000: My guess would be the default value and the ci overrides the same

rtnpro:
#dotests

rtnpro:
#dotests

rtnpro:
#dotests

PR #159, opened 2017-02-06
Installs docker-1.10.3-59.el7.centos.x86_64 than latest

navidshaikh:
Installs docker-1.10.3-59.el7.centos.x86_64 version in all container
  images than pulling the latest one

### DIFF
--- a/mail_service/Dockerfile.mailserv
+++ b/mail_service/Dockerfile.mailserv
@@ -1,8 +1,8 @@
 FROM registry.centos.org/centos/centos:latest
 
 RUN yum -y update && \
-    yum -y install docker && \
-    yum -y install wget mailx postfix rsyslog
+    yum -y install docker-1.10.3-59.el7.centos.x86_64 wget mailx postfix rsyslog && \
+    yum clean all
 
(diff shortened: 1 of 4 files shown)


kbsingh:
why are you hard wiring a specific version ?

navidshaikh:
> why are you hard wiring a specific version ?

@kbsingh : 
With latest version docker-1.12.5-14.el7.centos.x86_64, we are facing some regression.
I am testing a PR on devcloud setups, I have 2 setups, one with older while another with latest docker RPM installed.

With latest docker RPM installed, image push is failing (from cccp-build container) to OpenShift registry. While debugging, figured the docker push configuration is not present inside cccp-build container.
docker push configuration is shared inside cccp-build container via OpenShift using secrets. 
Logs shows secrets data is absent in image, thus image push failed and builds.

We are yet to find the fix for it, however the issue has blocked outstanding PRs. Unless we find the solution, we can keep using the docker 1.10 and move to latest docker once we are sure there are no regressions.

navidshaikh:
Closing the issue since the fix is merged in master with different PR.

PR #160, opened 2017-02-06
[Test] [do not merge] Docker check ci

navidshaikh:
Created a PR to check if CI is passing with EPEL issue fixed and using docker version 1.10.


mohammedzee1000:
Hmm it appears the error has changes so the docker this definetly one of the source of our current errors
@navidshaikh @rtnpro

navidshaikh:
#dotests

navidshaikh:
Closing the issue since the fix is in place.

PR #162, opened 2017-02-08
Fix the logic of git-path for paths starting with /.

mohammedzee1000:


jenkinsbuilder/cccp_index_reader.py
```diff
@@ -24,8 +24,10 @@ def projectify(
     new_project[0]['project']['name'] = appid
     new_project[0]['project']['git_url'] = giturl
     new_project[0]['project']['git_branch'] = gitbranch
-    new_project[0]['project']['rel_path'] = ('/' + gitpath) \
-        if (gitpath and not gitpath.startswith('/')) else '/'
+    rel_path = "/"
+    if gitpath and gitpath != "/":
```
navidshaikh: shouldn't either of condition suffice for the operation performed ?
like
`if gitpath != "/":` 
since if gitpath is `None`, the prior check before `and` does not make sense.

mohammedzee1000: Well this would assume that the user always passes a proper git path or ''. If a None is passed even by mistake, then:
```

In [1]: a=None

In [2]: rp = "/"

In [3]: a, rp
Out[3]: (None, '/')

In [4]: if a != "/":
   ...:     rp = a if a.startswith("/") else (rp + a)
   ...:     
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-fb79f32750b6> in <module>()
      1 if a != "/":
----> 2     rp = a if a.startswith("/") else (rp + a)
      3 

AttributeError: 'NoneType' object has no attribute 'startswith'
```

navidshaikh: Alright. I thought the input would be saner, but okay!

mohammedzee1000: Well i believe in trying to cover for as much of the cases as i can. So even assuming input is sane 99.9999% of the time i dont want to assume it always will be.

navidshaikh: well done!

PR #163, opened 2017-02-09
Sets up NFS client config on 4 nodes

navidshaikh:
This is to configure 4 nodes of CCCP deployment in CI as NFS clients.
  NFS server configurations can be configured via group variable.

### DIFF
--- a/provisions/group_vars/all
+++ b/provisions/group_vars/all
@@ -9,5 +9,3 @@ cccp_source_branch: master
 public_registry: registry.centos.org
 
 enable_epel: true
-
-test_nfs_share: 172.29.32.44:/nfsshare
--- a/provisions/group_vars/all
+++ b/provisions/group_vars/all
@@ -9,3 +9,5 @@ cccp_source_branch: master
 public_registry: registry.centos.org
 
 enable_epel: true
+
+test_nfs_share: 172.29.32.44:/nfsshare
--- a/ci/cccp_ci.py
+++ b/ci/cccp_ci.py
@@ -116,6 +116,7 @@ oc_slave={jenkins_slave_host}""").format(
     with open('hosts', 'w') as f:
         f.write(ansible_inventory)
 
+    print test_nfs_share
 
 def setup_controller(controller):
     # provision controller
--- a/ci/cccp_ci.py
+++ b/ci/cccp_ci.py
@@ -20,6 +20,7 @@ api = os.environ.get('API')
 ver = "7"
 arch = "x86_64"
 count = 4
+NFS_SHARE = "/nfsshare"
 
 repo_url = os.environ.get('ghprbAuthorRepoGitUrl') or \
     os.environ.get('GIT_URL')
@@ -67,6 +68,8 @@ def print_nodes():
 def generate_ansible_inventory(jenkins_master_host, jenkins_slave_host,
                                openshift_host, scanner_host):
 
+    test_nfs_share = scanner_host + ":" + NFS_SHARE
+
     ansible_inventory = ("""
 [all:children]
 jenkins_master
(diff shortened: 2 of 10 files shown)


navidshaikh:
#dotests

navidshaikh:
#dotests

provisions/group_vars/all
```diff
@@ -9,3 +9,5 @@ cccp_source_branch: master
 public_registry: registry.centos.org
 
 enable_epel: true
+
+test_nfs_share: 172.29.32.44:/nfsshare
```
navidshaikh: This is NFS share created in devcloud serving the range of clients in devcloud. ( @bamachrn created it)

navidshaikh:
#dotests

dharmit:
This is what I've been doing for NFS mounts on my DevCloud setup:

```bash
$ onevm list | awk '{print $1}' # this prints only VM IDs
ID
611
894
895
896
897

$ for i in `seq 894 897`; do myssh $i '( mkdir /srv/pipeline-logs )'; done
$ for i in `seq 894 897`; do myssh $i '( mount -t nfs 172.29.32.44:/nfsshare /srv/pipeline-logs )'; done
```

So, as far as I can see, we need to do only two things on the slaves in DevCloud:
- create a directory
- mount the NFS share

Installation of packages can be skipped AFAICT.

provisions/group_vars/all
rtnpro: @navidshaikh CI nodes cannot access devcloud. So, we need to setup another CI node, if it's needed.

navidshaikh: @rtnpro ACK!

navidshaikh:
> Installation of packages can be skipped AFAICT.
@dharmit : done.

rtnpro:
It seems like you are having an error in Ansible syntax:

```
ERROR! 'when' is not a valid attribute for a Play

The error appears to have been in '/root/container-pipeline-service/provisions/vagrant.yml': line 15, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Configure NFS server on scanner node for CI
  ^ here

Build failed: Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added 'n64.dusty.ci.centos.org,172.19.2.128' (ECDSA) to the list of known hosts.
 [WARNING]: Optional dependency 'cryptography' raised an exception, falling
back to 'Crypto'
[DEPRECATION WARNING]: Instead of sudo/sudo_user, use become/become_user and 
make sure become_method is 'sudo' (default).
This feature will be removed in a 
future release. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
ERROR! 'when' is not a valid attribute for a Play

The error appears to have been in '/root/container-pipeline-service/provisions/vagrant.yml': line 15, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Configure NFS server on scanner node for CI
  ^ here
```

navidshaikh:
> It seems like you are having an error in Ansible syntax:

```
ERROR! 'when' is not a valid attribute for a Play
```

Yes, updated the PR. I was using `when` in a play.

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

bamachrn:
How are we mounting the NFS mount points for CI and dev? I know we did this by adding entry in fstab for prod. we need the mount points in all the nodes roght?
@navidshaikh

navidshaikh:
> How are we mounting the NFS mount points for CI and dev? I know we did this by adding entry in fstab for prod. 

The PR should how we are configuring the NFS mounts in CI.  For dev env, we can configure the NFS mounts with `test=True`.

> we need the mount points in all the nodes roght?
Yes!

bamachrn:
LGTM

PR #172, opened 2017-02-22
Fixes list index out of range error

navidshaikh:
Fixes parsing openshift stdout message for build name


mohammedzee1000:
@rtnpro @bamachrn we should be good here if you can review :)

PR #174, opened 2017-02-23
Fix list index out of range issue in delivery worker

navidshaikh:
With openshift v1.2.1 , the output on stdout has changed.
  Earlier we were using openshift v1.3, and code was written according to it.
  With fixes needed for CI, we started using openshift v1.2.1, and this code change
  is needed to make code work with openshift v1.2.1


mohammedzee1000:
Nice this should work weather we use 1.2.1 or 1.3.1 of openshift.

Issue #180, opened 2017-03-01
Change build logs' extension


dharmit:
Same as #179 except for build logs instead of linter logs

navidshaikh:
Closing it with https://github.com/CentOS/container-pipeline-service/pull/181

PR #181, opened 2017-03-01
Modifies linter and build logs extention to .txt

### LINKED ISSUE #179: Change Dockerfile linter logs' extension
Dockerfile linter logs get stored on NFS with `.log` extension. When this location is accessed via web browser, it immediately prompts to save the logs.

Instead, we can change the extension to `.txt` to prevent such pop-up and view the logs in browser itself.

cc: @navidshaikh

navidshaikh:
Fixes #179 #180

  linter and build logs filename has .log extension, browser does not
  render them but starts the download of the file. This commit updates
  extension to .txt , with NFS configured for logs, users are notified
  over email with links to all the logs, with .txt extension, users can
  check the logs right in brower tab than downloading them before opening.

### DIFF
--- a/beanstalk_worker/constants.py
+++ b/beanstalk_worker/constants.py
@@ -16,7 +16,7 @@ SCANNERS_RESULTFILE = {
             "RPMVerify_scanner_results.json"]
 }
 
-LINTER_RESULTFILE = "linter_results.log"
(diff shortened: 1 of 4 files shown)


dharmit:
LGTM

bamachrn:
LGTM 👍

PR #183, opened 2017-03-01
Exports failed build logs on NFS

navidshaikh:
Earlier we used to attach build logs in failure email to user.
  Now, we export the failed build logs on NFS disk and sends link
  to author of image.


mail_service/worker_notify_user.py
```diff
@@ -7,24 +7,28 @@
 bs = beanstalkc.Connection(host="172.17.0.1")
 bs.watch("notify_user")
 
+LOGS_DIR = "/srv/pipeline-logs/"
+LOGS_URL_BASE = "https://registry.centos.org/pipeline-logs/"
```
mohammedzee1000: I have noticed these constants being declared explicitly in multiple places. Would it not be a good idea to put it in one place and import it everywhere instead?

dharmit: +1 to what @mohammedzee1000 said. We already have both these variables defined [here](https://github.com/CentOS/container-pipeline-service/blob/master/beanstalk_worker/constants.py). Please reuse it.

navidshaikh: This script goes in a mail-server container, while `constant.py` is a different chroot altogether.

navidshaikh: See https://github.com/CentOS/container-pipeline-service/blob/master/mail_service/Dockerfile.mailserv#L12

mohammedzee1000: Ok

beanstalk_worker/worker_start_build.py
```diff
@@ -48,20 +48,13 @@ def run_command(command):
         logger.log(level=logging.CRITICAL, msg=e.message)
         return e.message
 
-def notify_build_failure(namespace, notify_email, log_command):
+def notify_build_failure(namespace, notify_email, build_logs_file):
     msg_details = {}
     msg_details['action'] = 'notify_user'
-    msg_details['subject'] = "FAILED: Container-build failed " + namespace
```
dharmit: We're removing `subject` part from the message here. Are we taking care of it somewhere else? Meaning, will the email being sent out to the user have a subject?

beanstalk_worker/worker_start_build.py
```diff
@@ -143,17 +136,20 @@ def start_build(job_details):
...
+            logs = "Could not retrieve build logs."
+            logger.log(level=logging.CRITICAL, msg=logs)
         else:
             logger.log(level=logging.INFO,
                        msg="Writing build logs to NFS share..")
+        finally:
+            # Export logs on disk in either case
             export_build_logs(logs, build_logs_file)
 
         is_complete = run_command(status_command).find('Complete')
```
dharmit: What is `run_command` returning? `is_complete` variable makes me feel it should be a True/False scenario. But `< 0` check makes me think I'm wrong. Also, `run command` function has no comments on what it's doing! It's a function that's executing some command but unclear as to what it's returning.

mail_service/worker_notify_user.py
```diff
@@ -156,6 +160,25 @@ def notify_user_with_linter_results(job_info):
             job.delete()
             continue
 
+    if "build_failed" in job_info:
+        if job_info["build_failed"]:
```
dharmit: What is happening here? We're first checking if a key `build_failed` exists in `job_info` dictionary. And then we're checking if it's set to `True`? Can we just do the `True`/`False` check?

dharmit: I don't see an `else` clause to think there's some other situation being tackled here!

mail_service/worker_notify_user.py
```diff
@@ -156,6 +160,25 @@ def notify_user_with_linter_results(job_info):
...
+            print "==> Build with id= %s failed, sending failure email." \
+                % job_id
+
+            logs_url = job_info["build_logs_file"].replace(
+                LOGS_DIR, LOGS_URL_BASE)
+
+            subject = "FAILED: Container build failed " + job_info["namespace"]
```
dharmit: Is my subject question earlier in the review being addressed by this line?

navidshaikh: Yes,

beanstalk_worker/worker_start_build.py
dharmit: Resolved over stand-up

mohammedzee1000:
Possible flake rerunning tests to be sure
#dotests

mohammedzee1000:
#dotests

navidshaikh:
#dotests

navidshaikh:
hold on for merge, I am verifying the PR.

navidshaikh:
> hold on for merge, I am verifying the PR.

verified the PR on devcloud setup

bamachrn:
LGTM 👍

PR #184, opened 2017-03-02
Fixed project naming for desired tag with dots.

mohammedzee1000:
When desired tag has dots, the project name also has the same
as it is appid-jobid-desired tag. This fails project creation
as openshift project names cannot have dots.
Added logic to relplace the same wherever project name was
being used.

### DIFF
--- a/beanstalk_worker/worker_start_build.py
+++ b/beanstalk_worker/worker_start_build.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import beanstalkc
+import binascii
 import json
 from subprocess import Popen
 from subprocess import PIPE
@@ -9,7 +10,6 @@ import time
 import logging
 import sys
 import os
-import base64
 
 bs = beanstalkc.Connection(host="BEANSTALK_SERVER")
 bs.watch("start_build")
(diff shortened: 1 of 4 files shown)


mohammedzee1000:
#dotests

server/run-test.sh
```diff
@@ -35,8 +35,8 @@ docker pull ${FULL_FROM} || jumpto sendstatusmail
 
 _ "Re-tagging tested image (${FULL_FROM} -> ${FULL_TO})"
 docker tag ${FULL_FROM} ${FULL_TO} || jumpto sendstatusmail
-
-NS=${APPID}-${JOBID}-${DESIRED_TAG}
+printf "${APPID}${JOBID}${DESIRED_TAG}" | base64
```
navidshaikh: why this repetition ?

mohammedzee1000: ah that was from test but it should still be helpful to see the hash on screen

mohammedzee1000:
Wow it passed

mohammedzee1000:
@navidshaikh @dharmit if you guys can review this PR it would be good.

Once we start using hashes we might no longer have to worry about some of the other characters we check for as well in appid and jobid fields. The entire notion of replacing the characters came from the fact that they could not be part of openshift project names. 

Note : Openshift project names can be got by doing -

`$ printf "<jenkins-project-name>" | xxd -p `

beanstalk_worker/worker_start_build.py
```diff
@@ -89,7 +90,10 @@ def start_build(job_details):
         appid = job_details['appid']
         jobid = job_details['jobid']
         desired_tag = job_details['desired_tag']
-        namespace = appid + "-" + jobid + "-" + desired_tag
+        pre_namespace = str(appid) + "-" + str(jobid) + "-" + str(desired_tag)
+        namespace = hexlify(pre_namespace)
+        debug_log("Openshift project namespace hexlified from {0} to {1} can be hash can be reproduced with xxd tool"
```
navidshaikh: *{1}, hash can be reproduced [..]

typo there

mohammedzee1000: Thanks will fix

mohammedzee1000:
Ok appears to be working fine.

PR #186, opened 2017-03-03
Removed p.wait as it can cause deadlock with piped Popen.

mohammedzee1000:
Refer https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait
This happens in especially large builds.


dharmit:
LGTM

mohammedzee1000:
@kbsingh @dharmit @navidshaikh can we get this merged

PR #187, opened 2017-03-07
Optimizes Dockerfiles for images used on openshift node

navidshaikh:
Uses registry.centos.org repository URL for centos base image
   Adds yum clean all after all install steps
   Adds maintainer for the Dockerfiles


bamachrn:
LGTM 👍

PR #197, opened 2017-03-14
Track RPM repo for changes

rtnpro:
- [x] Script to check RPM repo for updates
- [x] Integrate with fedmsg to publish messages on package changes


src/tracking/config.py
```diff
@@ -0,0 +1,12 @@
+upstreams = {
```
mohammedzee1000: We should probably have this listed in DB somewhere instead of directly mentioning it here. These can be default but we need to allow user to specify their own third party rpm repositories if they want to.

rtnpro: That is the plan, but it will need some work and careful thought. So, we'll get this in, in a different PR. I plan to scan the containers and save their enabled repos to db. Then, a modified version of this script goes and fetches the update/change data.

mohammedzee1000: Sounds good

mohammedzee1000:
@rtnpro adding a few comments would be a good idea. While i can make sense of whats going on here, it would still be helpful. If we cant do it in this PR, then we should do it a separate one and that is ok.

src/tracking/management/commands/triggerbuilds.py
```diff
@@ -0,0 +1,52 @@
...
+            except IndexError:
+                continue
+            if (timezone.now() - last_updated).seconds < (
+                    settings.CONTAINER_BUILD_TRIGGER_DELAY):
+                logger.info('Waiting for package change burst to complete')
+                time.sleep(settings.CONTAINER_BUILD_TRIGGER_DELAY)
+                continue
+            to_build = set()
+            qs = ContainerImage.objects.filter(to_build=True)
+            for container in qs:
```
mohammedzee1000: Does this work well with multiple levels of dependency?

mohammedzee1000:
I think it would be a good idea to go for generalization at some point. Because except for the bit that figures out which rpms got updated and which containers have them, everything else will remain the same irrespective of package type.

src/tracking/management/commands/triggerbuilds.py
rtnpro: Yes

rtnpro:
#dotests-debug

rtnpro:
#dotests-debug

Issue #196, opened 2017-03-14
Builder does not accept Git URL with .git at the end


bamachrn:
builder is not running if there is a .git in the end of git url. To handle this we had removed .git from the urls in the end. But that does not work for gitlab. reason behind this is, builder does a cd to the directory after cloning. And if there is a .git in the end it tries to cd into the .git directory which is not a git repo.

PR #202, opened 2017-03-17
if there is .git extion in git url remove it while entering directory

bamachrn:
fixes https://github.com/CentOS/container-pipeline-service/issues/196


server/run.sh
```diff
@@ -17,6 +17,11 @@ git clone $SOURCE_REPOSITORY_URL||jumpto sendstatusmail
 
 
 dirname=${SOURCE_REPOSITORY_URL##*/}
+if [[ "$dirname" == *.git ]]
```
dharmit: Tangential comment here. `dirname` is a shell command provided by `coreutils` package. Maybe we should use a different variable name?

PR #203, opened 2017-03-17
Updated clone repo to fail  gitlab urls if they dont end with .git

mohammedzee1000:


mohammedzee1000:
Should be pulled in immediately after https://github.com/CentOS/container-pipeline-service/pull/202

PR #212, opened 2017-03-23
Hot-fix to delete a build job even in case of failure

dharmit:
For time being, delete the job from tube if it fails to build an image. This is to ensure that end user is not bombarded with few hundred e-mails in matter of hours.


bamachrn:
@dharmit yes, we would be putting checks for the try catch in start_build function for more detail exception handling.
LGTM, waiting for CI to complete

PR #214, opened 2017-03-24
Update to jenkins 2.5.1

rtnpro:
Updating Jenkins version to 2.5.1 to work around slow installation on Jenkins.


rtnpro:
#dotests-debug

provisions/roles/jenkins/master/tasks/ci.yml
```diff
@@ -17,11 +17,17 @@
   when: test
   sudo: yes
 
+- name: wait for 10 seconds
+  pause:
+    seconds: 10
+
```
navidshaikh: This wait is to let Jenkins create test job or something else? If yes, we are already doing 50 retries with 10 seconds delay. Please add comment with reason we are waiting for 10 seconds.

rtnpro:
#dotests-debug

rtnpro:
#dotests

rtnpro:
#dotests

provisions/roles/jenkins/master/tasks/ci.yml
rtnpro: @navidshaikh I was testing out something with delays. It has now been removed.

bamachrn:
LGTM, this will minimize the deployment time

PR #217, opened 2017-03-27
Added container-status in CC for recording user emails

bamachrn:
* Added container-status-report@centos.org in CC for all the mails sent from pipeline
* This will record all the mails sent from pipeline to user


bamachrn:
#dotests

bamachrn:
#dotests

PR #224, opened 2017-04-10
Fixes NFS server logs directory write permission config for CI

navidshaikh:
Adds more option in NFS server /etc/exports config file to allow
  other (jenkins) user to write to NFS. This configuration is only
  applicable for dev environement and CI, for prod, it is configured
  separately. This replicates the prod config though.


navidshaikh:
@bamachrn @arrfab : PTAL

PR #225, opened 2017-04-11
Adding unit tests for container pipeline service

mohammedzee1000:


jenkinsbuilder/test.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def test_01_main_creates_jjb_file_correct_data(self):
+        index_location = path.dirname(path.realpath(__file__)) + "/test_index"
+        data = cccp_index_reader.main(index_location, mock=True)
+        self.assertTrue(path.exists(data[0]["dump_file"]))
+
+    def test_02_main_file_data_matches_info(self):
+        print "=" * 40
+        print "Test if project data in file matches the information in index file"
+        print "=" * 40
```
dharmit: Below can be replaced by a single function:

```python
print "=" * 40
print "<some-message>"
print "=" * 40
```
Would come handy when writing more tests in future.

dharmit: I see you've done that already for `server/test.py` :smile:

server/test.py
```diff
@@ -0,0 +1,120 @@
...
+        self._test_script_path = None
+        self._delivery_script_path = None
+        touch_file(self._target_file)
+
+        if cccp_yml_data:
+            touch_file(self._cccp_mock)
+            with open(self._cccp_mock, "w") as f:
+                yaml.dump(cccp_yml_data, f)
+
+            if "build-script" in cccp_yml_data and cccp_yml_data["build-script"]:
```
dharmit: IMO, `if "build-script" in cccp_yml_data and cccp_yml_data["build-script"]` would yield same as `if cccp_yml_data.get("build-script")`

jenkinsbuilder/test.py
mohammedzee1000: Yea as i find more common stuff, i will be refactoring into a util of sorts :)

server/test.py
mohammedzee1000: Agreed, there is a lot of refactoring that needs to be done here.

server/cccp_reader.py
```diff
@@ -24,20 +24,20 @@ def main():
     if(key == "test-skip"):
       test_skip = cccp_yml["test-skip"]
 
-    if(key == "build-script"):
+    if(key == "build-script" and cccp_yml["build-script"]):
```
bamachrn: this is raise exception of key not found in cccp_yml, if it is not defined and stop the loop. may be we should put this in try catch block? similar for the below keys also.

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -110,15 +112,25 @@ def main(indexdlocation):
...
+                            proc = subprocess.Popen(myargs,
+                                                    stdout=subprocess.PIPE)
+                            proc.communicate()
 
                     finally:
                         print "Removing {}".format(t)
                         # shutil.rmtree(t)
-
+                        if mock:
+                            return ret_data
```
rtnpro: :+1: for adding ``if mock``. Consider using stubs, if needed.

I think that the function should spit the same data as output when called from test or from application code. It depends on the callee to decide what to do with the return data. It'd be great, not to add 'mocking' logic into the code being tested. This will add unnecessary complexity. It'd be great to break a big function into smaller ones which takes some data structure (dict or list) input and returns some output on which you can do some assertion on it.

server/cccp_reader.py
mohammedzee1000: This becomes an issue if someone passes None or empty string

mohammedzee1000: Understood, will apply this

jenkinsbuilder/cccp_index_reader.py
mohammedzee1000: I thought of using stubs, but then whats the point of stubbing the function under test, you ussually stub stuff that the function needs, but is in accessible due to n number of reasons, such as not having a fully functional jenkins or openshift for example.

rtnpro:
@mohammedzee1000 

Test cases look good.

A few points to consider:
- How do you plan ton integrate these tests with the CI?
- A few changes in tests scaffolding contradictory to our initial discussions
  - Let the tests lie in ``ci/tests`` for now, to be later moved to ``tests`` when we share common lib
  - the scaffolding for ``ci/tests`` could be:

    ```
    tests/
    |--- unit
         |--- test_01_feature_name
               |--- __init__.py
               |--- test_some_feature.py
               |--- data (to hold test data)
    |--- functional (similar structure as unit)
    ```
    IIRC ``container-pipeline-service`` directory is added in PYTHONPATH during CI

Once the unittests are in ``tests/unit`` dir, all we need to do to run them is ``nosetests ci/tests/unit``. Once that is done, we can wire up the test cases to CI build in a manner that:
- we run unittests first before getting nodes from CI
- only after successful run of unittests, we run functional tests

Does that make sense? Let me know if you have some question.

rtnpro:
#dotests-debug

rtnpro:
#dotests

rtnpro:
#dotests

rtnpro:
LGTM :+1: CI builds fail because of non related functional tests. I vote this PR to be merged.

navidshaikh:
#dotests

mohammedzee1000:
Unit tests passing locally

```
$ sudo nosetests -s ci/tests/unit_tests/
===========================================================================
Test if a project is projectified correctly if all parameters are populated
===========================================================================
.=====================================================
Test if main creates jjb_fileif correct data is given
=====================================================
creating: /tmp/tmpkFETc8
['jenkins-jobs', '--ignore-cache', 'update', 'project-defaults.yml:/tmp/tmpkFETc8/cccp_GENERATED.yaml']
Removing /tmp/tmpkFETc8
.==================================================================
Test if project data in file matches the information in index file
==================================================================
creating: /tmp/tmpmuCSKo
['jenkins-jobs', '--ignore-cache', 'update', 'project-defaults.yml:/tmp/tmpmuCSKo/cccp_GENERATED.yaml']
Removing /tmp/tmpmuCSKo
..===================================================
Test if echo scripts writes correct entry into file
===================================================
.=======================================================
Test if add_build_script adds correct RUN entry to file
=======================================================
.==================================================================================================
Test if the target file is generated correctly if all parameters are passed and test-skip is false
==================================================================================================
==> Getting current directory
==> Saving scripts for build process
==> scripts saved
.==========================================================
Test if correct target file generated if test skip is True
==========================================================
==> Getting current directory
==> Saving scripts for build process
==> scripts saved
.=================================================================
Test if target is generated correctly if no build script is given
=================================================================
==> Getting current directory
==> Saving scripts for build process
==> scripts saved
.===================================================================================
Test if target generation fails if no test script is given while test skip is false
===================================================================================
==> Getting current directory
==> Saving scripts for build process
.====================================================================
Test if target is generated correctly if no delivery script is given
====================================================================
==> Getting current directory
==> Saving scripts for build process
==> scripts saved
.================================================================
Test if target generation fails if job id is missing in cccp yml
================================================================
==> Getting current directory
==> Saving scripts for build process
.
----------------------------------------------------------------------
Ran 12 tests in 0.050s

OK

```

navidshaikh:
#dotests

navidshaikh:
#dotests-debug

navidshaikh:
#dotests-debug

navidshaikh:
@mohammedzee1000 : is this still relevant ?

navidshaikh:
#dotests-debug

mohammedzee1000:
No, its not. I will be setting up new tests and open a new PR.

PR #226, opened 2017-04-11
Adds line number and process number for code logging

navidshaikh:
This PR updates logging formatter config to include:
 pathname of the python script (earlier it was only name)
 process name
 line number in python script, string is logged from

### DIFF
--- a/beanstalk_worker/config.py
+++ b/beanstalk_worker/config.py
@@ -10,7 +10,7 @@ LOGGING_CONF = dict(
     level=LOGLEVEL,
     formatters=dict(
         bare={
-            "format": '[%(asctime)s %(name)s %(levelname)s] %(message)s'
+            "format": '[%(asctime)s] %(name)s p%(process)s %(lineno)d %(levelname)s - %(message)s'
         },
     ),
     handlers=dict(
--- a/mail_service/config.py
+++ b/mail_service/config.py
@@ -10,7 +10,7 @@ LOGGING_CONF = dict(
     level=LOGLEVEL,
     formatters=dict(
         bare={
-            "format": '[%(asctime)s %(name)s %(levelname)s] %(message)s'
+            "format": '[%(asctime)s] %(name)s p%(process)s %(lineno)d %(levelname)s - %(message)s'
         },
     ),
     handlers=dict(


navidshaikh:
@rtnpro PTAL

navidshaikh:
@dharmit : You asked for pasting, how would the resultant logging lines would like, here is a sample
```
[2017-04-13 12:54:29,461] p28701 {/opt/cccp-service/beanstalk_worker/worker_dockerfile_lint.py:86} INFO - Running Dockerfile Lint check
[2017-04-13 12:54:42,339] p28701 {/opt/cccp-service/beanstalk_worker/worker_dockerfile_lint.py:98} INFO - Dockerfile Lint check done. Exporting logs.
```

navidshaikh:
@bamachrn @mohammedzee1000 PTAL

rtnpro:
I am still not sure, if we really need this because:

- We already having unique logger names for each component of our application
- Having proper log messages makes unnecessary the need for line numbers
- Line numbers do not add much value to non error/exception logs, and having specified ``exc_info=True`` when calling the logging method, prints line numbers with traceback, when we need it
- having so much data on the log line will make it difficult to observe the main log messages

These are my 2 cents. You are free to agree/disagree with it.

navidshaikh:
> We already having unique logger names for each component of our application

The PR is not related to that though, it is not modifying that part.

> Having proper log messages makes unnecessary the need for line numbers

We have been having a lot of debugging recently, having enough information while doing debugging helps minimize devs time. 

>  having so much data on the log line will make it difficult to observe the main log messages

How ? just 2/3 characters at the start of the log message should not make it difficult. 

This PR also adds process numbers to log messages.

Since we have routed all the logs to a single file, having process number helps understand better if the log message is coming from container or normal / systemd process. For example, every logged message from (worker) containers will have process number 1 (p1).

I don't understand why we do not want to have verbose logging information. I think it helps debug better and understand the control flow throughout system.

navidshaikh:
I have removed the pathname of module for readability, since it was making log messages very long.

ping @rtnpro @bamachrn

navidshaikh:
@bamachrn : WDYT abou this ? I have updated the PR per suggestion, have removed the pathname of the module.

navidshaikh:
#dotests

bamachrn:
LGTM 👍

bamachrn:
#dotests

navidshaikh:
#dotests

rtnpro:
LGTM :

navidshaikh:
This is ready to merge.

PR #227, opened 2017-04-13
Adds missing centralize logging for dispatch worker

navidshaikh:
Enroutes all the messages generated from dispatch worker to
  centralize logging. It is deployed as a container.


rtnpro:
LGTM :+1:

PR #232, opened 2017-04-19
Adds cccp_index_repo variable under all vars in hosts.sample

navidshaikh:
Deployment failed complaining absense of cccp_index_repo variable.
 This was not the case earlier though.
 By adding the variable in all:vars section of hosts.sample file
 resolves the issue.
 This variable needs to be updated if doing testing deployment.
 The default value points to centos/container-index.gitt


navidshaikh:
@rtnpro ping

bamachrn:
LGTM

PR #233, opened 2017-04-21
Adds tags for all operations in ansible playbook

navidshaikh:
This PR aims to add tags in all the operations of ansible playbook.
 The reason for doing this is to have more control over deployment,
 in a way to run operations only on particular node.

 A subset of operations in existing playbook have tags, but this PR
 tags all the operations.

 How this will be useful: For example if deployment failed for some reason
 just before configuring scanner node, one can resume the installation by
 specifying scan tag while re-running the deployment.

### DIFF
--- a/provisions/roles/ci/tasks/main.yml
+++ b/provisions/roles/ci/tasks/main.yml
@@ -4,3 +4,4 @@
   tags:
     - test
     - ci-build-test-project
+    - ci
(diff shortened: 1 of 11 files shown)


navidshaikh:
#dotests

rtnpro:
Error:

```
TASK [openshift : include] *****************************************************
fatal: [n32.gusty.ci.centos.org]: FAILED! => {"failed": true, "reason": "the file_name '/root/container-pipeline-service/provisions/setup_mailservice.yml - openshift - mail_service' does not exist, or is not readable"}
	to retry, use: --limit @/root/container-pipeline-service/provisions/vagrant.retry
```

navidshaikh:
> Error:

@rtnpro : thanks for pointing it out, I had forgotten `tags:` keyword there, fixed it. 

This is ready to go @bamachrn

PR #235, opened 2017-04-24
Added registry.centos.org as defult registry

bamachrn:
* Images should be pulled in following sequence
1. internal registry
2. registry.centos.org
3. docker.io

### DIFF
--- a/provisions/hosts.sample
+++ b/provisions/hosts.sample
@@ -19,6 +19,7 @@ scanner_worker
 [all:vars]
 # update as needed
 public_registry=jenkins-slave
+intranet_registry=jenkins-slave:5000
 beanstalk_server=openshift
 rsync_ssh_opts=""
 
--- a/provisions/roles/openshift/tasks/config.yml
+++ b/provisions/roles/openshift/tasks/config.yml
@@ -3,13 +3,13 @@
   replace: >
     dest=/etc/sysconfig/docker
     regexp="^#*\s*INSECURE_REGISTRY=.*"
-    replace='INSECURE_REGISTRY="--insecure-registry 172.30.0.0/16 --insecure-registry {{ public_registry }}:5000"'
+    replace='INSECURE_REGISTRY="--insecure-registry 172.30.0.0/16 --insecure-registry {{ intranet_registry }}"'
 
 - name: Enable Docker registry
   replace: >
     dest=/etc/sysconfig/docker
     regexp="^#*\s*ADD_REGISTRY=.*"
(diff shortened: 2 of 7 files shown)


provisions/roles/openshift/tasks/config.yml
```diff
@@ -9,7 +9,7 @@
   replace: >
     dest=/etc/sysconfig/docker
     regexp="^#*\s*ADD_REGISTRY=.*"
-    replace='ADD_REGISTRY="--add-registry {{ public_registry }}:5000"'
+    replace='ADD_REGISTRY="--add-registry {{ public_registry }}:5000 --add-registry registry.centos.org"'
```
navidshaikh: >Images should be pulled in following sequence
> internal registry
> registry.centos.org
> docker.io

so {{public_registry}} here is not really public registry but internal registry ?

navidshaikh:
#dotests

provisions/roles/openshift/tasks/config.yml
bamachrn: yes, it is the public registry only but without going through the nginx

navidshaikh: okay

provisions/roles/repo_tracking/tasks/main.yml
```diff
@@ -94,7 +94,7 @@
 - name: Run postgres container
   docker_container:
       name: postgres
-      image: docker.io/centos/postgresql-95-centos7
+      image: registry.centos.org/sclo/postgresql-95-centos7
```
navidshaikh: shouldn't this be a separate PR @rtnpro ?

bamachrn: yes please

bamachrn:
@rtnpro I intentionally did not change the name from public_registry to intranet_registry, as there will be one more entry in internal_registry. This will create more confusions. value referred by intranet_registry is actually the pubilc registry.

provisions/hosts.sample
```diff
@@ -19,6 +19,7 @@ scanner_worker
 [all:vars]
 # update as needed
 public_registry=jenkins-slave
+intranet_registry=jenkins-slave:5000
```
bamachrn: why we are keeping both here the public_registry and intranet_registry?

navidshaikh:
> I intentionally did not change the name from public_registry to intranet_registry,

however the confusion stands ? IMO we change the names, we figure out the name when we want to add another one. For now, it would be good if we change the name.

bamachrn:
#dotests

PR #237, opened 2017-04-25
Added retries in case empty status is returned by openshift.

mohammedzee1000:
Sometimes, due to unknown reasons, openshift returns empty
string as status, which causes the workers to report a
build as failure, even though it might have succeeded.


mohammedzee1000:
@bamachrn @dharmit review please

bamachrn:
LGTM 👍

PR #238, opened 2017-04-25
Reduces number of characters in email subject

navidshaikh:
This changeset removes unnecessary information from email subject
  for success and failed builds. Now the build name in email subject
  is actual namespace/image_name of resulting image.

### DIFF
--- a/mail_service/worker_notify_user.py
+++ b/mail_service/worker_notify_user.py
@@ -24,8 +24,8 @@ BUILD_LOGS_FILENAME = "build_logs.txt"
 LINTER_STATUS_FILE = "linter_status.json"
 SCANNERS_STATUS_FILE = "scanners_status.json"
 
-SUCCESS_EMAIL_SUBJECT = "SUCCESS: Your container build: %s request is complete"
-FAILURE_EMAIL_SUBJECT = "FAILED: Container build failed: %s"
+SUCCESS_EMAIL_SUBJECT = "SUCCESS: Container build: %s is complete"
+FAILURE_EMAIL_SUBJECT = "FAILED: Container build: %s is failed"
 WEEKLY_EMAIL_SUBJECT = "Weekly scanning results for image: %s"
 
 EMAIL_HEADER = """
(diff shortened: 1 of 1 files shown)


mail_service/worker_notify_user.py
```diff
@@ -100,8 +100,8 @@ def __init__(self, job_info):
             logger.debug("Processing mail for failed build.")
             self.image_under_test = job_info.get("project_name")
 
-        self.project = self.job_info["namespace"] + "/" + \
-            self.job_info["job_name"]
+        self.project = self.job_info["namespace"].split("-")[0] + \
+            "/" + self.job_info["job_name"]
```
bamachrn: @navidshaikh may be we remove the job-name here and keep the name space as it is. this will make sure to keep uniqueness even for the image with different tag

mail_service/worker_notify_user.py
```diff
@@ -100,8 +100,8 @@ def __init__(self, job_info):
...
```
bamachrn: @navidshaikh may be we remove the job-name here and keep the name space as it is. this will make sure to keep uniqueness even for the image with different tag

navidshaikh: @bamachrn:  IMO, keeping the name as it will be for actual built image make more sense. Say user has added r.c.o in docker config, then user just need to look at the repository name and `docker pull`. 

@rtnpro @dharmit @mohammedzee1000 : WDYT ?

mohammedzee1000: Hmm i don't a user would and should be interested in the name we give to the project internally. What is he going to do with that information anyway. He is only interested in the name of the container with which he can pull the image, and that should be unique enough as it includes the full name, including tag.

navidshaikh: @mohammedzee1000 +1

navidshaikh:
#dotests

navidshaikh:
#dotests

bamachrn:
LGTM 👍

PR #239, opened 2017-04-25
Cleanup loggers for workers.

rtnpro:
- Removes unnecessary loggers
- Fixes log levels
- added some required INFO logs

### DIFF
--- a/beanstalk_worker/worker_start_build.py
+++ b/beanstalk_worker/worker_start_build.py
@@ -67,14 +67,13 @@ def export_build_logs(logs, destination):
         with open(destination, "w") as fin:
             fin.write(logs)
     except IOError as e:
-        logger.log(level=logging.CRITICAL,
-                   msg="Failed writing logs to %s" % destination)
-        logger.log(level=logging.CRITICAL, msg=str(e))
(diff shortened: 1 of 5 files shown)


beanstalk_worker/worker_start_build.py
```diff
@@ -194,16 +193,16 @@ def main():
...
                     bs.use('master_tube')
                     bs.put(json.dumps(job_details))
                     bs.use('failed_build')
                 else:
-                    logger.debug(str(job_details))
+                    logger.info('Starting build for job: %s' % job_details)
                     result = start_build(job_details)
             else:
-                logger.debug("No job found to process looping again")
+                logger.info("No job found to process looping again")
```
navidshaikh: shouldn't this be debug ? this will be the case most of the time.

rtnpro: That is debatable, and I am not sure about it, TBH. We need to have a discussion at some point of time to decide what to log in which level. Here's an interesting read: http://stackoverflow.com/questions/7839565/logging-levels-logback-rule-of-thumb-to-assign-log-levels

PR #240, opened 2017-04-25
Restart systemd services on failure

rtnpro:
This will ensure that the pipeline is functional even if there are service
failures


navidshaikh:
Excellent! This was needed.

rtnpro:
#dotests

PR #242, opened 2017-04-26
Uncomment NFS configuration flag in hosts.sample

navidshaikh:
we added a new line

	#setup_nfs=False

 this option needs to be updated every time a deployment
 is happening. Plus it has two valid options. For production,
 we need to make it False and True for dev environment.
 So, why to make it commented out ? If wee keep defaults
 for production, still it can be set to False (which it is)
 but without commented line.


bamachrn:
LGTM, merging

PR #243, opened 2017-04-27
Update README.md

dharmit:


rtnpro:
LGTM :+1: 

Some points to consider:

- Does the vagrant setup work? Our system has become much complicated and I don't think that it's possible to do a single node development on a laptop.

bamachrn:
lets remove this for now?

PR #246, opened 2017-05-02
Filters paths with known issues in rpm-verify scanner results

navidshaikh:
RPM verify is reporting issue in centos base image itself with
  filepath `/` shown as file attributes modification.
  All the layers built on top of centos base image is reporting same
  issue when scanner is ran on resulting image. The results are being
  sent to service users, and to avoid confusion we should filter this
  result from end result and get it fixed in base image. Once the issue
  is fixed in base image, we can update the scanner config to revert the
  change.

### DIFF
--- a/atomic_scanners/scanner-rpm-verify/rpm_verify_constants.py
+++ b/atomic_scanners/scanner-rpm-verify/rpm_verify_constants.py
@@ -1,0 +1,23 @@
+"""
+This is a constants file for rpm-verify atomic scanner
+In this file, one can mention the files to be excluded in end
+result of scanner OR Any other expected property configuration
+changes
+"""
+
+
+# Filter the paths you know the resulting image or base image itself
+# has issue about and need to be filtered
+# out since this is a known issue and it is in progress to get fixed.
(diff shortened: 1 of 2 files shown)


atomic_scanners/scanner-rpm-verify/rpm_verify_constants.py
```diff
@@ -0,0 +1,23 @@
...
+  "/",   # centos base image has issue with this filepath
+]
+
+
+# Filter filepaths starting with following directories listing,
+# since these paths are expected to be modified and should not
+# take into account
+
+FILTER_DIRS = [
+    "/var", "/run", "/media", "/mnt", "/tmp", "/proc", "/sys", "/boot"
```
bamachrn: are we filtering out all the apps in /var? I mean grep * /var -R or grep * /var.

navidshaikh: All the paths starting with `/var` , this is not a change, it's been around since we have scanner integrated. It just moved to constant file now.

PR #247, opened 2017-05-02
Distinguishes emails from prod or non prod environment

navidshaikh:
This changeset helps developer identify whether notification
  emails are coming from production or non production environment.

### DIFF
--- a/mail_service/worker_notify_user.py
+++ b/mail_service/worker_notify_user.py
@@ -125,8 +125,36 @@ class NotifyUser(object):
 
         return text.replace("\n", "\\n")
 
+    def update_subject_of_email(self, subject):
+        """
+        Mail server container is created with a environment variable
+        "PRODUCTION", its value is either True or False (Bool).
+        """
+        # Default case (if no env var found): It is production.
+        # PRODUCTION = "True" : no prepending
+        # PRODUCTION = "False" or any other value : prepend [test] in subject
+
+        # if var's value is "True" and even if the env variable is not found
+        # (since first priority is production)
+
+        if not os.environ.get("PRODUCTION", False) or \
+           os.environ.get("PRODUCTION") == "True":
+            # production
+
+            # no change
+            return subject
+        else:
+            # for any other value of "PRODUCTION" environment variable if set
+
+            # alter subject
+            return "[test] " + subject
+
+
     def send_email(self, subject, contents):
         "Sends email to user"
+
+        # process subject of email based on if it is production or not
+        subject = self.update_subject_of_email(subject)
 
         subprocess.call([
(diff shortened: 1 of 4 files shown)


mail_service/worker_notify_user.py
```diff
@@ -129,9 +129,37 @@ def _escape_text_(self, text):
...
+
+            # alter subject
+            return "[test] " + subject
+
+
     def send_email(self, subject, contents):
         "Sends email to user"
 
+        # process subject of email based on if it is production or not
+        self.update_subject_of_email(subject)
```
dharmit: I see you're returning string from `update_subject_of_email` function. But you're not using it in this line. You intended to do `subject = self.update_subject_of_email(subject)`?

provisions/hosts.sample
```diff
@@ -27,9 +27,14 @@ cccp_index_repo=https://github.com/centos/container-index.git
 
 ## dev environment options, update following to true to configure NFS
 setup_nfs=False
+
 # replace scanner_worker below with its FQDN / IP
 #test_nfs_share=scanner_worker:/nfsshare
 
+# for distinguishing emails from production or devcloud
+# keep it True for production and False for non production
+production=True
```
dharmit: Debatable or petty but, when we talked about this, I was suggesting to use something like `test=False`.

navidshaikh:
#dotests

provisions/hosts.sample
navidshaikh: We can utilize `test` variable for more relative operations, lets keep `production` variable name for now and we re-iterate as required.

navidshaikh:
#dotests

dharmit:
#dotests

provisions/roles/openshift/tasks/setup_mailservice.yml
```diff
@@ -20,10 +20,25 @@
...
+  when: not production
+  docker:
+    name: mail-server
+    state: restarted
+    image: mail-server
+    restart_policy: unless-stopped
+    volumes:
+      /srv/pipeline-logs/:/srv/pipeline-logs:rw
+    env:
+      PRODUCTION: False
```
bamachrn: for one variable value we are creating two blocks. there is no other way to do the same?

navidshaikh: Ansible allows boolean values for variables. Environment variables to be passed to container are bash environment variable, also the condition on which the env var value depends applies to sub operation of `Start Mail Service`. To avoid complexity, I simply put two different blocks.

navidshaikh:
#dotests

bamachrn:
LGTM 👍

PR #256, opened 2017-05-08
Update varchar len for package release to 100

rtnpro:
to accomodate for checksums in the release.


rtnpro:
@bamachrn This is also merge ready.

PR #258, opened 2017-05-08
Allow configuring logrotate policy for cccp.log.

rtnpro:


provisions/hosts.sample
```diff
@@ -34,3 +34,5 @@ setup_nfs=False
 [jenkins_master:vars]
 # update as needed
 oc_slave=jenkins-slave
+logrotate_maxsize=100M
+logrotate_rotate=5
```
bamachrn: @rtnpro why dont we keep it under all vars?

rtnpro: @bamachrn I did not put the vars in ``all:vars`` just to show that these belong to ``jenkins/master``  role and are used there only. However, it does not stop anyone to put it in ``all:vars`` in the inventory file.

rtnpro:
#dotests-debug

rtnpro:
#dotests-debug

rtnpro:
@bamachrn this is merge ready!

Issue #257, opened 2017-05-08
Make workers handle beanstalkd restart/failures gracefully


dharmit:
At the moment, if beanstalkd goes crazy i.e. restarts or fails, all the workers go nuts. All workers start logging messages related to connection failure and end up making the log file get really huge in size. Instead, we should make all workers handle the situation gracefully.

ping @rtnpro, you mentioned you'll work on this, right?

navidshaikh:
I think this is fixed when we refactored workers.

PR #261, opened 2017-05-09
Handle beanstalk socket errors in long running scripts

rtnpro:
This is an attempt to handle beanstalk connection errors in our long running scripts. We need this, because, currently on beanstalk connection errors, our services flood the log file with beanstalk connection error logs.

I have 2 solutions in mind for it's solution:

1. Log beanstalk error as critical and error out, and let ``systemd`` or ``docker`` restart the application with some delay, say **10 seconds**.
     **Pros**
     - Simple code in our services

    **Cons**
     - Docker, unlike systemd, does not support restart with delay

1. Handle beanstalk errors in our script, log it, and re create beanstalk connection with some delay.

    **Pros**
    - Will work with Docker

    **Cons**
    - Will lead to more code complexity

Currently, the pull request have implemented solution ``1``, and then I discovered that docker does not support restart with delay. A possible workaround could be running docker containers using systemd service files, which will need a little bit of work, but it seems cleaner.

However, I am not sure of what's the best method, so, I 'd like to hear your opinions on this.

### DIFF
--- a/beanstalk_worker/worker_dispatcher.py
+++ b/beanstalk_worker/worker_dispatcher.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import sys
 
 
 import beanstalkc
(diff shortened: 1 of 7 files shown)


dharmit:
@rtnpro When I tried doing it for scan worker (as mentioned in #257), I went with the 2nd approach. I added a delay in worker whenever it failed to connect to beanstalkd. And try to reconnect after few seconds. It worked fine. Agreed about the code complexity, but it ensured that the worker never exits at the time of beanstalkd issues.

rtnpro:
@dharmit Thanks for sharing your thoughts. We'll do it the way you took for scanner workers.

bamachrn:
https://github.com/CentOS/container-pipeline-service/pull/270 takes care of beanstalk issues

Issue #259, opened 2017-05-08
Pipeline capabilities scanner is reporting wrong name in scanner logs.


mohammedzee1000:
We have reports that scanner logs of container capabilities scanner is giving the wrong container name.
I have verified this and here are some samples of erroneous logs.

Container name : registry.centos.org:fabric8/openshift-nginx-tag:1.0.1
Log File : https://registry.centos.org/pipeline-logs/NDBjZGZkMGIyOG/container-capabilities-results.json
Name shown : c2bf.rdu2.centos.org:5000/mattermost/mattermost-team:3.8

Container Name : registry.centos.org:eclipse/centos-python35:latest
Log file : https://registry.centos.org/pipeline-logs/NDBjZGZkMGIyOG/container-capabilities-results.json
Name shown : c2bf.rdu2.centos.org:5000/mattermost/mattermost-team:3.8
We should get this fixed ASAP.

dharmit:
Issue is with the way we evaluate [`TEST_TAG`](https://github.com/CentOS/container-pipeline-service/blob/master/jenkinsbuilder/project-defaults.yml#L20).

Since for a weekly scan all the jobs get put onto the tube at almost same time, value for `TEST_TAG` for them is exactly same. This seems to be because same value till seconds for these jobs.

We can probably use `date +%N` instead to evaluate `TEST_TAG`. That ensures we get seconds since `1970-01-01 00:00:00 UTC` accurate up to nanoseconds.

dharmit:
Before the fix was applied:

```
Image simple/wget sent for weekly scan with data {'logs_dir': '/srv/pipeline-logs/OGVkYmFhZGE1NT', 'tag': 'latest', 'output_image': 'dev-32-132.lon1.centos.org:5000/simple/wget:latest', 'notify_email': 'shahdharmit@gmail.com', 'TEST_TAG': 'OGVkYmFhZGE1NT', 'action': 'start_scan', 'weekly': True, 'namespace': 'simple-wget-latest', 'job_name': 'wget', 'image_under_test': 'registry.centos.org:simple/wget:latest'}
[...]

PR #262, opened 2017-05-09
Remove duplicacy in TEST_TAG and do not add host info in scanner logs

dharmit:
As a part of #259 this PR fixes two items:

- `TEST_TAG` creation which uses `date` command's output is now accurate up to nanoseconds instead of a second. This is to ensure unique `TEST_TAG`s get created for weekly scan jobs. As a part of keeping things consistent, modified the same for all jobs through `projectdefaults.yml`

- Image name which also contains the host info is removed from container-capabilities-scanner's results since the email we send to developer already has that info in correct form.

### DIFF
--- a/jenkinsbuilder/project-defaults.yml
+++ b/jenkinsbuilder/project-defaults.yml
@@ -17,7 +17,7 @@
     builders:
         - shell: |
             export DOCKERFILE_DIR=`pwd`
-            TEST_TAG=`date | md5sum | base64 | head -c 14`
+            TEST_TAG=`date +%s%N | md5sum | base64 | head -c 14`
             LOGS_DIR=/srv/pipeline-logs/$TEST_TAG
             echo "Creating logs directory $LOGS_DIR"
             mkdir $LOGS_DIR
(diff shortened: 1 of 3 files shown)


jenkinsbuilder/project-defaults.yml
```diff
@@ -17,7 +17,7 @@
     builders:
         - shell: |
             export DOCKERFILE_DIR=`pwd`
-            TEST_TAG=`date | md5sum | base64 | head -c 14`
+            TEST_TAG=`date +%s%N | md5sum | base64 | head -c 14`
```
rtnpro: Shouldn't there be a ``space`` post ``+``. Just nitpicking ;) from PEP8 habits.

rtnpro:
LGTM :+1:

jenkinsbuilder/project-defaults.yml
dharmit: @rtnpro That's a shell command. Not a python thing. Doint that would lead to:
```bash
$ date + %s%N
date: extra operand ‘%s%N’
Try 'date --help' for more information.
```

PR #263, opened 2017-05-09
Dynamic debug loggers for project

rtnpro:
This PR is to implement the following:

- [x] Create debug log files for each project
- [x] Clean up debug logs if project build (including notifying user) went successfully

### DIFF
--- a/beanstalk_worker/worker_dispatcher.py
+++ b/beanstalk_worker/worker_dispatcher.py
@@ -1,11 +1,10 @@
 #!/usr/bin/env python
 
-import json
-import logging
-
-
 import beanstalkc
 import config
+import json
+import logging
+import os
 
 
 config.load_logger()
@@ -19,6 +18,9 @@ logger.info("Starting dispatcher")
 while True:
     job = bs.reserve()
     job_details = json.loads(job.body)
+    debug_logs_file = os.path.join(
+        job_details["logs_dir"], "service_debug.log")
+    dfh = config.DynamicFileHandler(logger, debug_logs_file)
     logger.info('Got job: %s' % job_details)
 
     logger.info("Retrieving job action")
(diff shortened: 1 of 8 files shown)


beanstalk_worker/worker_dockerfile_lint.py
```diff
@@ -137,11 +138,22 @@ def main():
         try:
             job = bs.reserve()
             job_data = json.loads(job.body)
+
+            debug_logs_file = os.path.join(
+                job_data["logs_dir"],
+                "service_debug.log"
```
mohammedzee1000: Why are we hard coding when it seems to be same in multiple places. Why not have it in config

beanstalk_worker/worker_dispatcher.py
```diff
@@ -56,3 +58,7 @@
 
     logger.debug("Deleting job: %s" % job_details)
     job.delete()
+
+    # remove handler from logging object
+    if 'dfh' in locals():
```
bamachrn: we are not putting any debug log, did not get the point to initialize and remove the handler without writing anything.

beanstalk_worker/worker_dockerfile_lint.py
bamachrn: I agree, linter and workers are not yet streamlined. so it would be hard to put them in a central point

bamachrn:
👍  for the idea to put the debug logs per build. we need reiterate the log levels for whole code base though

beanstalk_worker/worker_dispatcher.py
navidshaikh: If we set the debug log level, info messages will anyway be logged - is your question same or you want to ask something else ?

rtnpro:
LGTM :+1:

bamachrn:
LGTM :+1:

navidshaikh:
CI has passed and this is ready to merge.

PR #265, opened 2017-05-11
Removes unnecessary constant file from scanner code

navidshaikh:
There was an unnecessary constant file added to scanner code,
  which is uneasy to find by the actual scanner code since
  scanner is mounted as image and then ran inside another container.
  Paths needs to be added in environment variable with 2 containers
  involved. To reduce the complexity, compounded two files into one.


dharmit:
#dotests

dharmit:
LGTM

PR #270, opened 2017-05-17
Refactor pipeline workers

rtnpro:
This pull request does the following:

- [x] Refactors scaffolding of container pipeline service
   -  [x] Common lib for needed helpers
       - [x] Openshift
       - [x] Queue
       - [x] Settings
       - [x] Running system commands
- [x] Refactor build worker as a reference implementation with above refactor
- [x] Refactor delivery worker
- [x] Refactor linter worker
- [x] Refactor test worker
- [x] Refactor docker file lint worker
- [x] Fix provisioning


container_pipeline/workers/build.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def handle_build_failure(self, job):
+        """Handle build failure for job"""
+        self.queue.put(json.dumps(job), 'failed_build')
+        self.logger.info(
+            "Build is not successful putting it to failed build tube")
+        appid = job['appid']
+        jobid = job['jobid']
+        desired_tag = job['desired_tag']
+        project = appid + "/" + jobid + ":" + desired_tag
```
dharmit: Can we rather use `project = job.get('appid') + "/" +job.get('jobid') + ":" + desired_tag`. 

For two reasons:

- We don't seem to be using `appid` and `jobid` elsewhere in the function.
- `job['appid']` will throw exception if the key doesn't exist for some reason. `.get()` will just give a `None` value.

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,76 @@
...
+        build_logs_file = os.path.join(job['logs_dir'], 'build_logs.txt')
+        self.notify_build_failure(
+            get_job_name(job), job['notify_email'], build_logs_file,
+            project, job['jobid'], job['TEST_TAG'])
```
dharmit: The closing `)` should be on newline?

container_pipeline/lib/command.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+import subprocess
+
+
+def run(cmd):
```
dharmit: Can we name it `run_cmd` or `run_command` to separate it from `run` method we're having in workers as well?

container_pipeline/workers/build.py
dharmit: We don't want `None` value in either of those three either. Might also be worth ensuring none of them have `None` values.

container_pipeline/lib/openshift.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def get_build_status(self, project, build_id):
+        output = run(
+            'oc get --namespace {project} build/{build_id} {suffix} | '
+            'grep -v STATUS'.format(
+                project=project, build_id=build_id)
+        )
+        return output.split()[3]
+
+    def wait_for_build_status(self, project, build_id, status, retries=None,
```
bamachrn: We are not providing the value for retires from build? and also there is no certain number, it will take to get the build complete.

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,62 @@
...
+        data['project_name'] = project
+        data['job_name'] = jobid
+        data['TEST_TAG'] = test_tag
+        self.queue.put(json.dumps(data), 'master_tube')
+
+    def export_build_logs(self, logs, destination):
+        """"Write logs in given destination"""
+        # to take care if the logs directory is not created
+        if not os.path.exists(os.path.dirname(destination)):
+            os.makedirs(os.path.dirname(destination))
```
mohammedzee1000: We probably should, as a precaution if nothing else.

container_pipeline/lib/command.py
```diff
@@ -0,0 +1,16 @@
...
+    Runs a shell command.
+
+    :param cmd: Command to run
+    :type cmd: str
+
+    :return: Command output
+    :rtype: str
+    :raises: subprocess.CalledProcessError
+    """
+    return subprocess.check_output(cmd, shell=True)
```
dharmit: Do we have a strong reason to use `shell=True`? There's plethora of content on the Internet that suggests against using it. Or can we get rid of it in a safe manner?

container_pipeline/lib/default_settings.py
```diff
@@ -0,0 +1,68 @@
...
+        },
+    ),
+    handlers=dict(
+        console={
+            "class": "logging.StreamHandler",
+            "formatter": "bare",
+            "level": LOG_LEVEL,
+        },
+        log_to_file={
+            'level': 'DEBUG',
```
dharmit: Why aren't we using the variable `LOG_LEVEL` here?

container_pipeline/lib/default_settings.py
```diff
@@ -0,0 +1,68 @@
...
+        'mail-service': {
+            "level": "DEBUG",
+            "propagate": False,
+            "handlers": ["console", "log_to_file"],
+        },
+        'console': {
+            "level": "DEBUG",
+            "propagate": False,
+            "handlers": ["console"]
+        }
```
dharmit: Why aren't we using the variable `LOG_LEVEL` in this dictionary? About 7 occurrences here.

container_pipeline/lib/default_settings.py
```diff
@@ -0,0 +1,68 @@
...
+        log_to_file={
+            'level': 'DEBUG',
+            'class': 'logging.FileHandler',
+            'filename': LOG_PATH,
+            'mode': 'a+',
+            'formatter': 'bare',
+        }
+    ),
+    loggers={
+        'linter': {
```
dharmit: Should we use `dockerfile-linter` here instead? Support for other image formats might come up in future, right? cc: @bamachrn

container_pipeline/lib/openshift.py
```diff
@@ -0,0 +1,77 @@
...
+from container_pipeline.lib import settings
+
+from .lib.command import run
+
+
+class OpenshiftError(Exception):
+    pass
+
+
+class Openshift(object):
```
dharmit: For all the `oc` commands we're using here, is it plausible to use [this Python API](https://github.com/openshift/openshift-restclient-python) instead?

container_pipeline/lib/command.py
dharmit: If there's an exception in executing the `cmd`, does the caller need to handle the raised exception or should we have that logic here?

dharmit: I'd vote for handling here so that all the methods that use this function in any way get a consistent error message and avoid the scenario where individual workers' developers have their own implementation of handling such a scenario.

dharmit: For example, we have following code:
```python
def get_build_logs(self, project, build_id):
        output = run(
            'oc logs --namespace {project} build/{build_id} {suffix}'.format(
                project=project, build_id=build_id,
                suffix=self.oc_cmd_suffix))
        return output
```

What if `output` contains exception info and not expected output? If we handle it here and then subsequent methods just return the exception info till the top caller, top caller can take decision based on an `if..else` check. Correct me if I'm wrong. Nitpicking on this function since we expect it to be one of the most re-used ones across the code.

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self, logger=None):
+        self.queue = JobQueue(host=settings.BEANSTALKD_HOST,
+                              port=settings.BEANSTALKD_PORT)
+        self.logger = logger or logging.getLogger('console')
+        self.openshift = Openshift(logger=self.logger)
+        self.openshift.login()
+
+    def handle_job(self, job):
+        """Handle job"""
+        raise NotImplementedError
```
dharmit: We don't have an implementation of this exception yet in the refactor. Just FYI.

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,62 @@
...
+                             project, jobid, test_tag):
+        """Notify build failure"""
+        data = {}
+        data['action'] = 'notify_user'
+        data['namespace'] = namespace
+        data['build_status'] = False
+        data['notify_email'] = notify_email
+        data['build_logs_file'] = build_logs_file
+        data['project_name'] = project
+        data['job_name'] = jobid
```
dharmit: Can we try and use same key name and value variable on left and right? For eg:
```python
data['project_name'] = project_name
data['job_name'] = job_name
```
OR
```python
data['project'] = project
data['jobid'] = jobid
```
instead of current implementation? This is to avoid confusions. We've made mistakes in past owing to such implementations.

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,62 @@
...
+        while True:
+            job_obj = self.queue.deque()
+            job = json.loads(job_obj.body)
+            self.logger.info('Got job: %s' % job)
+            try:
+                self.handle_job(job)
+            except Exception as e:
+                self.logger.critical(
+                    e, extra={'locals': locals()}, exc_info=True)
+            self.queue.delete(job_obj)
```
dharmit: - This is deliberately not put into `finally` clause because we're doing retries, right? 
- Whether we fail to perform the requested job after configured number of retries **or** we succeed in performing it, the job's getting deleted, right?

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,76 @@
...
+from .base import BaseWorker
+
+
+class BuildWorker(BaseWorker):
+    """Build worker"""
+
+    def handle_job(self, job):
+        """Handle job"""
+        parent_build_running = False
+        parents = job.get('depends_on', '').split(',')
```
dharmit: In the container-index CI, we are having a check to ensure that `depends-on` isn't empty, right? @mohammedzee1000

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def handle_build_success(self, job):
+        """Handle build success for job"""
+        self.logger.debug("Build is successful going for next job")
+
+    def handle_build_failure(self, job):
+        """Handle build failure for job"""
+        self.queue.put(json.dumps(job), 'failed_build')
+        self.logger.info(
+            "Build is not successful putting it to failed build tube")
```
dharmit: The closing `)` should be on newline?

container_pipeline/lib/default_settings.py
dharmit: [Buildah](https://github.com/projectatomic/buildah) for example.

container_pipeline/lib/default_settings.py
```diff
@@ -0,0 +1,68 @@
+LOG_LEVEL = 'DEBUG'
```
rtnpro: :+1:

container_pipeline/lib/logging.py
```diff
@@ -0,0 +1,7 @@
+import logging
+
+from .config import LOGGING_CONF
```
rtnpro: half baked refactor. fixed!

container_pipeline/lib/openshift.py
```diff
@@ -0,0 +1,77 @@
+import logging
+import time
+
+from container_pipeline.lib import settings
+
+from .lib.command import run
```
rtnpro: ack :+1: will fix it.

container_pipeline/lib/openshift.py
```diff
@@ -0,0 +1,77 @@
...
+class Openshift(object):
+
+    def __init__(self, endpoint=None, config=None, cert=None, logger=None):
+        self.endpoint = endpoint or settings.OPENSHIFT_ENDPOINT
+        self.config = config or settings.OC_CONFIG
+        self.cert = cert or settings.OC_CERT
+        self.oc_cmd_suffix = (
+            '--config {config} --certificate-authority {cert}'.format(
+                config=self.config, cert=self.cert))
+        self.logger = logger or logging.getLogger('oc')
```
rtnpro: They got missed. I have added them.

container_pipeline/lib/settings.py
```diff
@@ -0,0 +1 @@
+from container_pipeline.lib.default_settings import *  # noqa
```
rtnpro: To make linters ignore errors/warnings. In this case, we're import ``*``, which is not usually OK, but it works here.

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+class BaseWorker(object):
+    """Base class for pipeline workers"""
+
+    def __init__(self, logger=None):
+        self.queue = JobQueue(host=settings.BEANSTALKD_HOST,
+                              port=settings.BEANSTALKD_PORT)
+        self.logger = logger or logging.getLogger('console')
+        self.openshift = Openshift(logger=self.logger)
+        self.openshift.login()
```
rtnpro: We'll remove it from the base class and put it in the workers where we need openshift.

container_pipeline/workers/build.py
rtnpro: @bamachrn said that we won't get ``None``  values anymore.

container_pipeline/lib/command.py
rtnpro: Yes. I will be updating it in an upcoming commit.

container_pipeline/lib/openshift.py
rtnpro: Fixed!

container_pipeline/lib/command.py
rtnpro: We use pipes in a number of commands, so, I believe that we'll need ``shell`` support. Also, it makes it easy to call the method. That said, I don't disagree with the potential misuse of it. However, our system does not have much user interaction from the outside world. The point of entry to our system is jenkins and the jobs are written by us. So, I don't see much room for exploit. The only variables that we have is the container-index params, which can be sanitized during linting the index.

However, this calls for a discussion.

rtnpro: 1. I'd prefer handing subprocess errors from ``run_cmd`` in the calling function, as the needs of the callers might vary. The callers are usually lib helpers we've written, which promise to abstract out the complexities for their users. But, the lib helpers need to know what has happened.

1. ``subprocess.check_output`` blows up if the exit code from the called command is non zero. So, we won't get any output if the command has error'd out. If some command errors out without a non zero exit code (which should not be the case), the caller can handle it.

container_pipeline/lib/default_settings.py
rtnpro: It's a typo. I will fix it/

container_pipeline/lib/default_settings.py
rtnpro: Yes, it makes sense to use ``dockerfile-linter``

container_pipeline/lib/default_settings.py
rtnpro: Good catch!

container_pipeline/lib/openshift.py
rtnpro: Python API will take more development time than using system commands. Most of the time, I have seen that libraries don't have 1-1 mapping with the command line. That said, it's good to have, and we can do it sometime. Since, we have abstracted out Openshift in a class, it will be easy to refactor without breaking other code.

container_pipeline/workers/base.py
rtnpro: This is done to force the subclass to define ``handle_job`` method, as it is a must for our workers to handle the job it receives.

container_pipeline/workers/base.py
rtnpro: :+1: legacy from older codebase.

container_pipeline/workers/base.py
rtnpro: yes

container_pipeline/workers/build.py
rtnpro: I am not sure, so I took my precaution.

container_pipeline/workers/build.py
rtnpro: Flake8/PEP8 lint does not complain.

container_pipeline/workers/build.py
dharmit: Ack! This validation needs to happen at an earlier stage. Few stages I can imagine are:
- when someone opens a PR on container-index, index CI should check it. cc: @mohammedzee1000 
- when Jenkins puts the data on to the tube, we should be sure that it won't put erroneous/incomplete data. This one has particularly bit us in past when `TEST_TAG` was missing and mail server went into a frenzy.

container_pipeline/lib/command.py
dharmit: Agree with the pipe usage in our code. One such command we're executing is:
```python
# TEST_TAG generation, unique per project                                  
task = subprocess.Popen(                                                
    "date +%s%N | md5sum | base64 | head -c 14",                           
    shell=True,                                                         
    stdout=subprocess.PIPE)                                             
TEST_TAG = task.stdout.read()
```
I'm replacing that `+%s%N` with `+%m%d%Y` so that we can have comparative output.
```python
In [5]: task = subprocess.Popen("date +%m%d%Y | md5sum | base64 | head -c 14", shell=True, stdout=subprocess.PIPE)

In [6]: TEST_TAG = task.stdout.read()

In [7]: TEST_TAG
Out[7]: 'OGQ0NDk5NjVmOD'
```
Without `shell=True`, it gets long and ugly.
```python
In [2]: date_cmd = subprocess.Popen(('date', '+%m%d%Y'), stdout=subprocess.PIPE)

In [3]: md5sum = subprocess.Popen(('md5sum'), stdin=date_cmd.stdout, stdout=subprocess.PIPE)

In [4]: base64_op = subprocess.Popen(('base64'), stdin=md5sum.stdout, stdout=subprocess.PIPE)

In [5]: headc14 = subprocess.check_output(('head', '-c', '14'), stdin=base64_op.stdout)

In [6]: date_cmd.communicate()
Out[6]: ('', None)

In [7]: headc14
Out[7]: 'OGQ0NDk5NjVmOD'
```
The longer approach also calls for some modification in `run_cmd` code, I guess. It should accept `**kwargs`, at least. That enables one to pass `shell=True` if absolutely needed (which can be discussed when someone's going to open such a PR :smile:).

I'm not in favor of or against any approach yet. But, this should help in discussions.

dharmit: Having said that, I do agree that our system's not going to have interaction with user provided scripts. It's always going to be workers that interact with it.

navidshaikh: > Having said that, I do agree that our system's not going to have interaction with user provided scripts. It's always going to be workers that interact with it.

+1 to that.

We should see the complexity involved. If its easy enough to avoid `shell=True`, why give any scope of exploit. Having said that, we should also consider the iteration model, where if it is not absolutely needed atm, we fix such things when more important things are fixed.

dharmit: @rtnpro As discussed, please make this function accept `**kwargs` so that if someone's trying to use `shell=True` in any worker, we can ensure its absolute necessity and then merge it.

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def build(self, job):
+        """Build job"""
+        namespace = get_job_name(job)
+        project = hashlib.sha224(namespace).hexdigest()
+
+        try:
+            self.openshift.login()
+            build_id = self.openshift.build(project, 'build')
+            if not build_id:
```
dharmit: When would we hit something like this where `if not build_id:` is satisfied? Just trying to understand. Because in this scenario, I don't see we are logging the reason for failure. Which in turn leaves admins and end user unaware of the cause of failure.

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,89 @@
...
+                return False
+        except OpenshiftError as e:
+            logger.error(e)
+            return False
+
+        Build(namespace).start()
```
dharmit: This will be connected to my earlier comment on the line about 
```python
is_build_running = Build(parent).is_running()
```
- Value of `namespace` for the [same index entry](https://github.com/CentOS/container-index/blob/master/index.d/pipeline-images.yml#L2-L11) will look like: `pipeline-images-cccp-build-latest`
- for the object of `utils.Build` class, values of `self.name` and `self._path` will be `pipeline-images-cccp-build-latest` and `/srv/pipeline-logs/pipeline-images-cccp-build-latest` respectively.
- when `start()` method is called, it will create a path `/srv/pipeline-logs/pipeline-images-cccp-build-latest`

Let's see its parent image `centos/centos:latest` from the `depends-on` field. [Here's its index entry](https://github.com/CentOS/container-index/blob/master/index.d/centos.yml#L2-L11).

When this image's build is started, the path it will create upon `Build(namespace).start()` will be: `/srv/pipeline-logs/centos-centos-latest`. But we are checking for `/srv/pipeline-logs/centos/centos:latest` file location in `is_build_running = Build(parent).is_running()`, right?

Unless I'm missing something, this looks a bug to me.

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,89 @@
...
+        build_status = self.openshift.wait_for_build_status(
+            project, build_id, 'Complete')
+        logs = self.openshift.get_build_logs(project, build_id)
```
dharmit: If `build_status` is failed, and there were empty retries, we're still going to be able to fetch build logs from openshift, right? Just checking. Because if not, this and the next two lines till `return build_status` don't serve much purpose. For example if OpenShift itself is down. (I know that's extremely pessimistic of me.) :smile:

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,89 @@
...
+        build_logs_file = os.path.join(job['logs_dir'], 'build_logs.txt')
+        self.export_build_logs(logs, build_logs_file)
+        return build_status
+
+    def handle_build_success(self, job):
+        """Handle build success for job"""
+        self.logger.debug("Build is successful going for next job")
```
dharmit: `job` parameter is provided to this function but it's not using it. 

The `run()` method of `Base` class already has the job info, right?

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,63 @@
...
+        data = {}
+        data['action'] = 'notify_user'
+        data['namespace'] = namespace
+        data['build_status'] = False
+        data['notify_email'] = notify_email
+        data['build_logs_file'] = build_logs_file
+        data['project_name'] = project
+        data['job_name'] = jobid
+        data['TEST_TAG'] = test_tag
+        self.queue.put(json.dumps(data), 'master_tube')
```
dharmit: If we're putting failed build data to `master_tube`, why are we creating the tube `failed_build`?

container_pipeline/lib/command.py
rtnpro: Fixed by fc70db431d9f6dd242b108beacc1269369e1405c

container_pipeline/lib/default_settings.py
rtnpro: Fixed by 2f682df463d5861ccf61681fa0fa1e2fa628ec35

container_pipeline/lib/default_settings.py
rtnpro: Fixed by 238e3b3892b353fc133cfa625ac95411746bb6b9

container_pipeline/lib/default_settings.py
rtnpro: We aren't using var ``LOG_LEVEL`` here because we wan't the loggers to generate all logs, and let the handlers decide which log levels to capture.

container_pipeline/lib/openshift.py
```diff
@@ -0,0 +1,132 @@
...
+        """Run build for a project"""
+        self.logger.debug('Run openshift project build: {}/{}'
+                          .format(build, project))
+        try:
+            output = run_cmd(
+                'oc --namespace {project} start-build {build} '
+                '{suffix}'.format(
+                    project=project, build=build, suffix=self.oc_cmd_suffix)
+            )
+            build_id = output.split('"')[-1].rstrip()
```
bamachrn: @dharmit this would yield the build number e.g. Build-1 or Test-1 or Delivery-1 based on the stages we are running this.

container_pipeline/workers/base.py
bamachrn: @rtnpro rather than creating the job details here. why don't we pass the job details received from the previous phase. And add necessary data. this will make sure we never miss some data in the job details
```bash
def notify_build_failure(self, job_detail,build_log_file)
{
   job_detail['action'] = 'notify_user'
   data['build_logs_file'] = build_logs_file
  self.queue.put(json.dumps(data), 'master_tube')
}
```
because rest of the data is already there in the job_details.

container_pipeline/workers/build.py
bamachrn: should we keep all the depends on values in parents? because it can have only one FROM. and all are not build time dependency also. I believe we need to discuss around this more.

container_pipeline/workers/build.py
```diff
@@ -0,0 +1,89 @@
...
+        super(BuildWorker, self).__init__(logger, sub, pub)
+        self.openshift = Openshift(logger=self.logger)
+
+    def handle_job(self, job):
+        """Handle job"""
+        parent_build_running = False
+        parents = job.get('depends_on', '').split(',')
+        parents_in_build = []
+        for parent in parents:
+            is_build_running = Build(parent).is_running()
```
bamachrn: also if we can may be we can use the TEST_TAG here. as it is available through out the system, and we are creating dir for this in the beginning. Though I have a doubt how we can fetch the status for a non-running previous build

dharmit: > also if we can may be we can use the TEST_TAG here. as it is available through out the system, and we are creating dir for this in the beginning.

@bamachrn you're right about `TEST_TAG`'s availability across the system. However, a directory created using `TEST_TAG` doesn't get deleted from NFS. That's probably not a good way to judge if a parent image is building *right now*. This is probably a use case which can leverage redis/rabbitmq, maybe?

container_pipeline/workers/base.py
rtnpro: Fixed by 292e0bc606edfaf761d0160309a59d628e9deb05

container_pipeline/lib/openshift.py
```diff
@@ -0,0 +1,132 @@
...
+        password = password or self.password
+        try:
+            self.logger.debug(
+                run_cmd(
+                    'oc login {endpoint} -u {user} -p {password} {suffix}'
+                    .format(
+                        endpoint=self.endpoint, user=user,
+                        password=password, suffix=self.oc_cmd_suffix
+                    ))
+            )
```
rtnpro: This will not print the command we run, just the output

container_pipeline/lib/openshift.py
rtnpro: @dharmit I will document this once I have a working development instance. Devcloud is down. @bamachrn if you could add the expected output in comments above this line, it'd be great.

@bamachrn one question. If the output is ``Build-1``, ``Test-1``, why are we splitting with `"`?

container_pipeline/lib/openshift.py
```diff
@@ -0,0 +1,132 @@
...
+                    project=project, build_id=build_id,
+                    suffix=self.oc_cmd_suffix))
+            self.logger.debug('Build logs for project build: {}/{}\n{}'.format(
+                project, build_id, output))
+        except subprocess.CalledProcessError:
+            self.logger.error(
+                'Could not retrieve build logs for project build: '
+                '{}/{}'.format(project, build_id))
+            output = 'Could not retrieve build logs'
+        return output
```
rtnpro: This is fixed!

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,63 @@
+import json
```
rtnpro: Got that!

rtnpro: Fixed by 26556c453d3849b3c959a940487cde5689615285

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,63 @@
...
+        while True:
+            job_obj = self.queue.get()
+            job = json.loads(job_obj.body)
+            self.logger.info('Got job: %s' % job)
+            try:
+                self.handle_job(job)
+            except Exception as e:
+                self.logger.error(
+                    'Error in handling job: {}\nJob details: {}'.format(
+                        e, job), extra={'locals': locals()}, exc_info=True)
```
rtnpro: @dharmit This is being taken care by https://github.com/CentOS/container-pipeline-service/pull/263. Also, I'd like to add a column in our logs which contains some unique job id, so that we can filter the logs. However, that we'll solve in another PR.

container_pipeline/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        + str(job_details['desired_tag'])
+
+
+class Build:
+    """
+    Track image build status in the pipeline
+
+    In future, we can collate a lot of duplicate code from workers
+    to manage and track build, thereby providing a clean interface
+    to manage builds and prevent duplication of code.
```
rtnpro: ok

rtnpro: Fixed by 27e897334e9b1fcb68ca9a7abce10a9b5a650e9f

container_pipeline/utils.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    def __init__(self, name, datadir='/srv/pipeline-logs'):
+        self.name = name
+        self._path = os.path.join(datadir, name)
+
+    def is_running(self):
```
rtnpro: Yes, it's confusing. So, but essentially, the output of this method means whether the pipeline build is running or not. So, in that way, it makes sense from an API consumer point of view.

container_pipeline/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        if os.path.exists(self._path):
+            return True
+        else:
+            return False
+
+    def start(self):
+        """Mark build as running"""
+        f = open(self._path, 'a')
+        os.utime(self._path, None)
+        f.close()
```
rtnpro: @dharmit You need the path to exist to do a ``os.utime(self._path, None)``.

navidshaikh:
@rtnpro : PR #263 is merged, you might want to incorporate the changes (adding dynamic file handler) here ?

container_pipeline/workers/build.py
rtnpro: yeah!

container_pipeline/workers/build.py
rtnpro: if for some reason, openshift fails to create a build and not return a build id. We had this scenario covered in the earlier codebase: https://github.com/CentOS/container-pipeline-service/blob/master/beanstalk_worker/worker_start_build.py#L128

container_pipeline/utils.py
dharmit: Ack, that was silly. I missed that this the `start` method.

container_pipeline/workers/build.py
rtnpro: We just use namespace value from job details, which gets generated from ``container_pipeline.utils.get_job_name``. And, there's no ``:`` in the generated name.

container_pipeline/workers/build.py
rtnpro: We might be able to fetch build logs, but there's no guarantee. We have faced scenarios, when openshift does not return any logs. @bamachrn ^^ ? We still need ``build_status`` because we use it's value in ``handle_job`` to call the ``handle_build_success`` or ``handle_build_failure`` method.

container_pipeline/workers/build.py
rtnpro: But ``run()`` does not know or care about if the job build was successful or not. We'll use ``job`` in this function during log refactor. Currently, it was just a copy paste from the existing code.

container_pipeline/lib/default_settings.py
rtnpro: Solved!

container_pipeline/lib/openshift.py
rtnpro: Solved!

container_pipeline/lib/openshift.py
rtnpro: Add a message along with the log: ``Logging in to openshift``

rtnpro: Solved!

container_pipeline/lib/openshift.py
rtnpro: @bamachrn will add comments of sample o/p/

container_pipeline/utils.py
rtnpro: Add docs in the method.

rtnpro: Fixed!

container_pipeline/utils.py
rtnpro: Fixed!

container_pipeline/workers/build.py
rtnpro: Fixed!

container_pipeline/workers/build.py
rtnpro: Will fix it later. Issue created: https://github.com/CentOS/container-pipeline-service/issues/276

container_pipeline/workers/build.py
rtnpro: Fixed!

container_pipeline/workers/base.py
dharmit: created issue #278 to track this.

dharmit:
LGTM for scaffolding! Thanks @rtnpro.

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,74 @@
...
+                              sub=sub, pub=pub, logger=self.logger)
+
+    def handle_job(self, job):
+        """
+        This method is called to process job data from task queue.
+        """
+        raise NotImplementedError
+
+    def notify_build_failure(self, namespace, notify_email, build_logs_file,
+                             project_name, job_name, test_tag):
```
dharmit: @rtnpro we need to make this method more generic and less inclined towards build worker

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,74 @@
...
+        data['build_status'] = False
+        data['notify_email'] = notify_email
+        data['build_logs_file'] = build_logs_file
+        data['project_name'] = project_name
+        data['job_name'] = job_name
+        data['TEST_TAG'] = test_tag
+        self.queue.put(json.dumps(data), 'master_tube')
+        self.logger.debug('Notify build failure: {}'.format(data))
+
+    def export_build_logs(self, logs, destination):
```
dharmit: Again, need to make this more generic and less inclined towards build worker.

rtnpro:
@dharmit https://github.com/CentOS/container-pipeline-service/pull/270/commits/695eb3c66493603fd30bc585aa760585c8175eda should address the concerns you raised above.

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,65 @@
...
+    def notify(self, data):
+        """
+        This method queues user notifications to be processed by the
+        mail service worker. Customize as needed.
+        """
+        self.queue.put(json.dumps(data), 'master_tube')
+
+    def export_logs(self, logs, destination):
+        """"Write logs in given destination"""
+        self.logger.debug('Writing build logs to NFS share')
```
navidshaikh: I want to re-use this method to
  - write dockerfile to given specified location to run linter
  - Write dockerfile linter result
  - Write dockerfile linter execution status
  - Write scanner results 
  - write scanner executions status
  - write delivery worker logs

Can we make it generic export logs method and remove the debug line or make it to
`"writing logs to NFS share path: %s"`

container_pipeline/workers/base.py
```diff
@@ -0,0 +1,65 @@
...
+        This method is called to process job data from task queue.
+        """
+        raise NotImplementedError
+
+    def notify(self, data):
+        """
+        This method queues user notifications to be processed by the
+        mail service worker. Customize as needed.
+        """
+        self.queue.put(json.dumps(data), 'master_tube')
```
navidshaikh: if the method use-case is to notify user,  can we explicitly do here
`data["action"] = 'notify_user'` and then put the details on master tube?

container_pipeline/workers/base.py
dharmit: :+1: maybe let it take one more argument so that in the debug message, we can write what worker's logs we're writing.

container_pipeline/workers/base.py
dharmit: We modified this method to be very simple and put the data to `master_tube`. Whatever worker's calling it has to ensure that the `data` being provided to this method has appropriate value for `action` key.

container_pipeline/workers/base.py
navidshaikh: Keeping the debug message argument optional with a default (Generic) message in method itself.

@rtnpro : WDYT ?

container_pipeline/lib/command.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+import subprocess
+
+
+def run_cmd(cmd, shell=False):
```
dharmit: I'm not able to comprehend having a dedicated file for just one function. Not that it's bad. But we also have `utils.py` which is having a number of utilities that we need to reuse across various files. Why haven't we included this function as a part of `utils.py`?

container_pipeline/workers/base.py
rtnpro: We should not take such ``debug`` message arguments, rather log it in the caller method appropriately. :+1: to making this class more generic.

rtnpro:
#dotests

rtnpro:
#dotests

rtnpro:
#dotests

rtnpro:
#dotests-debug

rtnpro:
#dotests-debug

rtnpro:
#dotests-debug

rtnpro:
@dharmit we have already made the base worker class more generic. Can you mark your review as complete?

container_pipeline/lib/command.py
rtnpro: A few reasons for this is:

- ``run_cmd`` will expand in the future, where we can have streaming output, etc.
- the structure of the ``lib`` pkg is that each of it's module contains a single functionality
- also, ``lib`` is something which is totally unaware of the pipeline, however, things in utils are kind of pipeline specific

I hope that this answers your questions.

dharmit: Sounds good. Thanks!

dharmit:
Approved the workers part which I had reviewed.

mohammedzee1000:
Nice CI green :)

bamachrn:
LGTM from the workers point of view at least 👍

PR #274, opened 2017-05-23
Updated index ci to fail on appid-filename mismatch.

mohammedzee1000:
This is to enforce the matching of filename with appid in the project. Until now it was a warning, but moving forward, we need the match.


navidshaikh:
If you can add details about changeset in the PR description, that would be helpful to understand the issue this PR was created for.

rtnpro:
LGTM :+1:

rtnpro:
@mohammedzee1000 can you remove the ``print("Verified")`` statements. It's not required. Do this in some other PR.

bamachrn:
LGTM 👍

PR #280, opened 2017-05-24
Remove scanner results from /var/lib/atomic

navidshaikh:
This change cleans the original results saved in /var/lib/atomic
  directory after scanner execution on given image. Once scanner
  finishes execution, results are loaded from default location and written
  on NFS disk. Thus there was unnecessary copy of result in /var/lib/atomic
  directory. This change removes the extra copy from /var/lib/atomic
  thus saves scanner node from running out of disk space.


navidshaikh:
Closing the PR since it is raised from wrong branch.

PR #282, opened 2017-05-26
Retains per build log file for failed builds

navidshaikh:
This changeset changes the way we are handling per build service
debug logs. Earlier we removed the logs irrespective of build status,
whether build succeeds, fails, we will remove the per build log file.
Now, service build log is kept for:

- Weekly:
  - Service failure / excpetion etc
- Normal builds:
  - Service failure /exception etc
  - Build failure


rtnpro:
LGTM :

bamachrn:
LGTM 👍

PR #283, opened 2017-06-06
Adds missing hardware related files in scanner expected deviations

navidshaikh:
There are a bunch of files reported as missing from systemd package
  in scanner rpm verify result for base image itself.
  `$ rpm -Va`  <-- running this command inside container.
  We found out that image is optimized to remove unnecessary files
  from container image and thus are the files reported as missing.
  The chageset adds the list of files in scanner expected deviations
  list so that it wont show up in the result.


bamachrn:
LGTM :+1:

PR #286, opened 2017-06-15
Refactor repo tracking feature

rtnpro:
This migrates the code for repo tracking to ``container_pipeline`` dir.

- [x] Code refactor/migration
- [x] Ground work for db integration with rest of codebase


Vagrantfile
```diff
@@ -2,65 +2,55 @@
 # vi: set ft=ruby :
 VAGRANTFILE_API_VERSION = "2"
 
-ALLINONE = (ENV['ALLINONE'] || 0).to_i
+PROD = (ENV['PROD'] || 0).to_i
```
mohammedzee1000: Whats with this change, do you mean prod like setup

rtnpro: It's for vagrant. Initially, we were having ``ALLINONE`` and multinode setup, but it seems that openshift and scanner do not run well together. So, I moved from a single node setup to 2 node setup, for development, and a multinode setup with 3-4 nodes for a prod like setup.

My choice of using ``PROD`` is mostly because of lack of good words.

bamachrn:
LGTM 👍

PR #287, opened 2017-06-16
Fix serilaized builds infinite loop

### LINKED ISSUE #271: Issue while parent job is building
In our current service while building a project in builder we first check if it has a depends on. It has a parent container image in its dependency chain. The parent container is either not there or building, then builds checks for the dependency and  keeps the build back in start_build tube.

```bash
if parent_build_running:
         logger.info('Parents in build: %s, pushing job: %s back '
                            'to queue' % (parents_in_build, job_details))
         bs.use('master_tube')
         bs.put(json.dumps(job_details))
```

Now the scenario is, if suppose image1 is in the build worker and image1 depends on image2. image2 is not there in registry, and image1 request is the only one request in build tube. Then image1 checks the dependency and fails and puts the request goes back in the tube after failing. This goes on infinitely and fills up the logs which causes the issue.

rtnpro:
Fixes #271 

Changes

- Add delay between build job retries, when parent build/test/scan/delivery is in progress

Benefits:
- This will help us rate limit the number of times we process/retry to process a job for build
- This will prevent generating unnecessary logs when processing a build job

### DIFF
--- a/container_pipeline/workers/base.py
+++ b/container_pipeline/workers/base.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import time
 
 from container_pipeline.lib import settings
 from container_pipeline.lib.queue import JobQueue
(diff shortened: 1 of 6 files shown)


container_pipeline/workers/base.py
```diff
@@ -49,17 +50,26 @@ def run(self):
...
+                debug_logs_file = os.path.join(
+                    job['logs_dir'], settings.SERVICE_LOGFILE)
+                # Run dfh.clean() to clean log files if no error is
+                # encountered in post delivering build report mails to user
+                dfh = DynamicFileHandler(self.logger, debug_logs_file)
+                self.logger.info('Got job: {}'.format(job))
+                try:
+                    self.handle_job(job)
+                except Exception as e:
+                    self.logger.error(
```
bamachrn: do we want to log only if there is a change in build state (if it is waiting, then only log if it starts building)?

bamachrn:
otherwise LGTM 👍

container_pipeline/workers/base.py
rtnpro: We are not tracking any change in build state, it kind of complicates things. This approach is far simpler, you either process the job or not. The thing is that we don't log or process the job if it's being retried and the current time has not passed the delay window.

rtnpro:
#dotests-debug

PR #288, opened 2017-06-16
Changes logging for build worker

dharmit:


container_pipeline/workers/build.py
```diff
@@ -35,11 +35,11 @@ def handle_job(self, job):
                 is_build_running
 
         if parent_build_running:
-            self.logger.info('Parents in build: {}, pushing job: {} back '
+            self.logger.debug('Parents in build: {}, pushing job: {} back '
```
navidshaikh: I think this message should stay as `info`. This message has information and reason for pushing the job back to queue. And I think it is a clue for log viewer to understand why job is pushed back to queue.

dharmit: The way I've differentiated between `info` and `debug` is that whenever we're storing some dynamic data into the logs, it goes under `debug`. In this case, we're logging the data about `parents_in_build` and `job` which we get to know dynamically only. Because, as a user I'm not really going to make much from that data. But an admin/dev would find that information helpful.

bamachrn: agree with the idea

bamachrn:
LGTM 👍

PR #289, opened 2017-06-16
Changes logging for delivery worker

dharmit:


container_pipeline/workers/delivery.py
```diff
@@ -83,7 +83,7 @@ def handle_delivery_failure(self, job):
         and requests to notify the user about failure to deliver
         """
         self.queue.put(json.dumps(job))
-        self.logger.info(
+        self.logger.warning(
             "Delivery is not successful putting it to failed delivery tube")
```
navidshaikh: service failing at delivery phase is a point of concern and should be logged as critical. Its not often we fail at delivery.

bamachrn:
LGTM 👍

PR #290, opened 2017-06-16
Changes logging for dockerfile lint worker

dharmit:


bamachrn:
👍  may be we can mark critical only when something is breaking the pipeline

PR #296, opened 2017-06-29
Refactor build project

rtnpro:


bamachrn:
@rtnpro start_build.py can we change this name to some thing like create_project.py or something like that. it is actually not starting the build. it is creating the project in openshift and updating the template. build worker actually starts the build.

container_pipeline/start_build.py
```diff
@@ -0,0 +1,75 @@
...
+    is_openshift_good = True
+
+    openshift = Openshift(logger=logger)
+    try:
+        openshift.login("test-admin", "test")
+        openshift.create(project)
+        openshift.clean_project(project)
+    except OpenshiftError:
+        is_openshift_good = False
+        return 1
```
navidshaikh: We are returning here.

container_pipeline/start_build.py
```diff
@@ -0,0 +1,75 @@
...
+            'APPID': appid,
+            'JOBID': jobid,
+            'REPO_BUILD_PATH': repo_build_path,
+            'TARGET_FILE': target_file,
+            'NOTIFY_EMAIL': notify_email,
+            'DESIRED_TAG': desired_tag,
+            'TEST_TAG': test_tag})
+    except OpenshiftError:
+        is_openshift_good = False
+        return 1
```
navidshaikh: And we are returning here too.

container_pipeline/start_build.py
```diff
@@ -0,0 +1,75 @@
...
+
+    if(is_openshift_good):
```
navidshaikh: in what scenario `is_openshift_good` would be `False`? all the other places we are setting it to `False` and returning.

bamachrn: If there is some error in project creation of openshift. we should send build requests to workers.

container_pipeline/start_build.py
```diff
@@ -0,0 +1,75 @@
...
+import sys
+
+import container_pipeline.utils as utils
+from container_pipeline.lib import settings
+from container_pipeline.lib.log import load_logger
+from container_pipeline.lib.openshift import Openshift, OpenshiftError
+from container_pipeline.lib.queue import JobQueue
+
+
+def create_project(appid, jobid, repo_url, repo_branch, repo_build_path, target_file, notify_email, desired_tag, depends_on, test_tag):
```
rtnpro: This function should either return ``True``/``False`` rather than ``0`` and ``1``.

rtnpro: Or it can just pass the exception to be handled by the caller.

container_pipeline/start_build.py
```diff
@@ -0,0 +1,75 @@
...
+    job_name = utils.get_job_name({
+        'appid': appid, 'jobid': jobid, 'desired_tag': desired_tag})
+    project = utils.get_job_hash(job_name)
+    is_openshift_good = True
+
+    openshift = Openshift(logger=logger)
+    try:
```
rtnpro: Redundant ``try...except`` blocks. Prefer raising exceptions and exiting the function as early as possible.

container_pipeline/start_build.py
```diff
@@ -0,0 +1,75 @@
...
+            'repo_build_path': repo_build_path,
+            'target_file': target_file,
+            'notify_email': notify_email,
+            'depends_on': depends_on,
+            'logs_dir': '/srv/pipeline-logs/{}'.format(test_tag),
+            'TEST_TAG': test_tag
+        }))
+        return 0
+    else:
+        logger.critical("Jenkins is not able to setup openshift project")
```
rtnpro: This should not be a ``critical`` log as the system is not breaking here. We can make it ``error``.

bamachrn: but here we are not building a job requested by user, so we should be knowing it. otherwise user and us both will be wondering what happened with that job.

container_pipeline/pipeline.py
```diff
@@ -0,0 +1,78 @@
...
+            'APPID': appid,
+            'JOBID': jobid,
+            'REPO_BUILD_PATH': repo_build_path,
+            'TARGET_FILE': target_file,
+            'NOTIFY_EMAIL': notify_email,
+            'DESIRED_TAG': desired_tag,
+            'TEST_TAG': test_tag})
+        is_openshift_good = True
+    except OpenshiftError:
+        raise
```
rtnpro: You can remove the ``try...except`` block here altogether.

container_pipeline/pipeline.py
```diff
@@ -0,0 +1,78 @@
...
+
+    if(is_openshift_good):
```
rtnpro: ``(...)`` is not needed

PR #300, opened 2017-07-06
Updates OS test-admin user policy to admin user

navidshaikh:
Earlier we had `test-admin` user configured with `view` role.
  This changeset configures `test-admin` user with `admin` role.
  Also, the provisioning bits are modified to not ignore errors.
  We need to make sure, during the deployment policy is added to user
  and not ignore errors. This is must since the user needs to have proper access.


bamachrn:
LGTM 👍

navidshaikh:
#dotests-debug

navidshaikh:
#dotests

PR #303, opened 2017-07-07
[WIP] Add provision roles to clean up before deployment.

rtnpro:
This adds (will add) provision roles to cleanup or bring down services in the pipeline before deployment so that we can avoid race conditions between deployment and earlier services running.


bamachrn:
this has been taken care of by https://github.com/CentOS/container-pipeline-service/pull/320

PR #306, opened 2017-07-12
Not to leave dependson as blank for null value

bamachrn:
if depends on is left as blank bnuild project fails to build it as it misses a command line argument.
so putting `none` in place of blank

### DIFF
--- a/jenkinsbuilder/cccp_index_reader.py
+++ b/jenkinsbuilder/cccp_index_reader.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python
+"""Create jenkins jobs from the container-index."""
 import os
-import yaml
-import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


jenkinsbuilder/cccp_index_reader.py
```diff
@@ -88,6 +88,9 @@ def get_projects_from_index(indexdlocation):
                             dependson_job = dependson_job.replace(
                                 ':', '-').replace('/', '-')
 
+                        if dependson_job == '':
+                            dependson_job = 'none'
```
navidshaikh: `none` or `None` ?

bamachrn: intentionally I kept it as string `none` do not want to use any system keyword

PR #307, opened 2017-07-12
Not to repeat failed builds

bamachrn:
Changed actions for failed builds, so that they do not repeat. if the action remains same it keeps on going to the same loop

### DIFF
--- a/container_pipeline/workers/build.py
+++ b/container_pipeline/workers/build.py
@@ -4,12 +4,12 @@ import logging
 import os
 import time
 
-from container_pipeline.utils import Build, get_job_name, get_project_name, \
-    get_job_hash
+from container_pipeline.lib import settings
 from container_pipeline.lib.log import load_logger
 from container_pipeline.lib.openshift import Openshift, OpenshiftError
+from container_pipeline.utils import (Build, get_job_hash, get_job_name,
+                                      get_project_name)
 from container_pipeline.workers.base import BaseWorker
-from container_pipeline.lib import settings
 
 
 class BuildWorker(BaseWorker):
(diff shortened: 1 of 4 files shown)


container_pipeline/workers/build.py
```diff
@@ -87,7 +87,9 @@ def handle_build_success(self, job):
 
     def handle_build_failure(self, job):
         """Handle build failure for job"""
-        self.queue.put(json.dumps(job))
+        job.pop('action', None)
+        job['action'] = "build_failure"
```
navidshaikh: you can just over ride it, line no:90 is not needed.

container_pipeline/workers/delivery.py
```diff
@@ -84,7 +84,9 @@ def handle_delivery_failure(self, job):
         Puts the job back to the delivery tube for later attempt at delivery
         and requests to notify the user about failure to deliver
         """
-        self.queue.put(json.dumps(job))
+        job.pop('action', None)
+        job['action'] = "delivery_failure"
```
navidshaikh: same here as well, just over ride `action` field, line no: 87 is not needed.

container_pipeline/workers/worker_test.py
```diff
@@ -51,7 +51,9 @@ def handle_test_success(self, job):
 
     def handle_test_failure(self, job):
         """Handle test failure for job"""
-        self.queue.put(json.dumps(job))
+        job.pop('action', None)
+        job['action'] = "test_failure"
```
navidshaikh: Same here too, remove line no:54

container_pipeline/workers/build.py
bamachrn: these were intentional. do not want get into great risk of repeating mails. rather I would love to keep the extra line. how do you think?

navidshaikh: What I think is, the operations in two lines can be clubbed into one line only and it is not giving risk of repeating emails.

navidshaikh: But its okay, if an extra line of operation is clearly indicating the goal of operation.

PR #308, opened 2017-07-12
Fixes linter command invocation via library run_cmd

navidshaikh:
Library `run_cmd` function takes string or list.
  One needs to pass list if caller wants `shell=True`, else just pass
  command as a string. This changeset fixes linter invocation, it does not
  need `shell=True`, and it was earlier using list to send command.
  Fixed it to send invocation commmand as string instead of list.


bamachrn:
LGTM 👍 I would ask @dharmit to put his comment here

PR #309, opened 2017-07-12
Added project existing check before creating it

bamachrn:
It checks for the project in openshift first. if the project is existing it does not create it.


bamachrn:
#dotests

PR #311, opened 2017-07-13
Syncs jenkins jobs with container index entries

### LINKED ISSUE #310: Jenkins does not sync jobs in change of container-index
If there is any update on entries for container-index jenkins does not reflect it properly.
**scenario:**
  cccp-index job in jenkins creates and updates all the jobs based on the current entries in entries.
So, if there is any update on the job parameters in index it gets reflected properly.

But if one index entry is removed from the container-index, jenkins does not clear up the job from its build config. As a result of it jenkins keeps on building/tracking old job configs which are deleted. Because of this images in registry.centos.org also remains for the deleted job.

**How to reproduce:**
     1. create a container index entry with more than one project in it.
     2. run the cccp-index job in jenkins, it will create entries for the projects.
     3. now delete one entry from the container-index
     4. run the cccp-index job again
    this will not delete the entry which is removed from container-index.

navidshaikh:
Fixes #310
  This changeset adds option to `jenkins-job` command to delete old
  project while updating it per index.

### DIFF
--- a/jenkinsbuilder/cccp_index_reader.py
+++ b/jenkinsbuilder/cccp_index_reader.py
@@ -185,7 +185,7 @@ def delete_stale_projects_on_jenkins(stale_projects):
     for project in stale_projects:
         myargs = ["jenkins-jobs", "delete", project]
         # print either output or error
-        out, error = run_command(myargs)
+        _, error = run_command(myargs)
         if error:
             logger.critical("Failed to delete project %s. Exiting..", project)
             logger.critical(sys.exc_info()[0])
@@ -228,7 +228,12 @@ def main(indexdlocation):
                       ':'.join(
                           [jjb_defaults_file, generated_filename])
                       ]
-            run_command(myargs)
+            _, error = run_command(myargs)
+            if error:
+                logger.critical("Error %s running command %s" % (
+                    error, str(myargs)))
+                logger.critical("Project details: %s ", str(project))
+                exit(1)
 
             new_projects_names.append(
                 str(project[0]["project"]["appid"]) + "-" +
@@ -238,6 +243,7 @@ def main(indexdlocation):
         except Exception as e:
             logger.critical("Error updating jenkins job via file %s",
                             generated_filename)
+            logger.critical("Project details: %s", str(project))
             logger.critical(str(e))
             # if jenkins job update fails, the cccp-index job should fail
             logger.critical(sys.exc_info()[0])
--- a/jenkinsbuilder/cccp_index_reader.py
+++ b/jenkinsbuilder/cccp_index_reader.py
@@ -122,6 +122,10 @@ def export_new_project_names(projects_names):
     with open(projects_list, "w") as fin:
         fin.write(projects_names)
 
+    # change the mod of file so that jenkins user can edit it
+    print "Changing file permission 0777 of file %s" % projects_list
+    run_command(["chmod", "0777", projects_list])
+
 
 def get_old_project_list():
     """
--- a/jenkinsbuilder/cccp_index_reader.py
+++ b/jenkinsbuilder/cccp_index_reader.py
@@ -1,12 +1,16 @@
 #!/usr/bin/env python
 """Create jenkins jobs from the container-index."""
+
+import logging
 import os
 import subprocess
 import sys
 import tempfile
-from glob import glob
-
 import yaml
+
+from container_pipeline.lib.log import load_logger
+
+from glob import glob
 
 jjb_defaults_file = 'project-defaults.yml'
 
@@ -108,9 +112,10 @@ def get_projects_from_index(indexdlocation):
                                 dependson, notifyemail, desiredtag)
                         )
                     except Exception as e:
-                        print "Failed to projectify %s" % str(project)
-                        print str(e)
-                        print sys.exc_info()[0]
+                        logger.critical("Failed to projectify %s" %
+                                        str(project))
+                        logger.critical(str(e))
+                        logger.critical(sys.exc_info()[0])
                         raise
     return projects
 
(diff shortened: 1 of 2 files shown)


navidshaikh:
#dotests-debug

bamachrn:
@navidshaikh how do you mean by deleting old jobs? if we delete them it losses track of last build and last change set. so every time cccp-index job runs all the projects will get build. could you please explain a bit. I not seeing how this will delete the jobs which are not there in index.

navidshaikh:
#dotests-debug

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -100,16 +108,87 @@ def get_projects_from_index(indexdlocation):
...
 
+def export_new_project_names(projects_names):
+    """
+    Exports the name of project
+    """
+    # opens the file defined at the start of this script
+    projects_names = "\n".join(projects_names)
+    with open(projects_list, "w") as fin:
+        fin.write(projects_names)
+
```
bamachrn: how about putting this in try.. catch block

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -118,23 +197,41 @@ def main(indexdlocation):
...
+    print stale_projects
+
+    # delete stale entries at jenkins
+    delete_stale_projects_on_jenkins(stale_projects)
+
+    # delete stale entries at openshift
+    delete_stale_projects_on_openshift()
 
+    # export the current projects_list in file
+    export_new_project_names(new_projects_names)
```
bamachrn: this part also in try catch

navidshaikh: that would be try catching which is already taken care in function. Doing try catch at function itself.

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -100,16 +108,93 @@ def get_projects_from_index(indexdlocation):
...
+def delete_stale_projects_on_jenkins(stale_projects):
+    """
+    This function deletes the stale entries at jenkins that
+    are no longer valid. If there is any issue/exception, it
+    bypasses by printing the error.
+    """
+    for project in stale_projects:
+        myargs = ["jenkins-jobs", "delete", project]
+        # print either output or error
+        print run_command(myargs)
```
dharmit: What if there's an error in executing the `run_command`? We don't seem to be doing any exception handling here.

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -100,16 +108,93 @@ def get_projects_from_index(indexdlocation):
...
+
+
+def delete_stale_projects_on_openshift():
+    """
+    Stale projects found on jenkins need to be deleted
+    accordingly on OpenShift as well. This function maps
+    the project names from jenkins to openshift and deletes them.
+    (see delete_stale_projects_on_jenkins function)
+    """
+    pass
```
dharmit: This function is getting called and all it's got is `pass`. We either need the implementation or a `#TODO` comment to indicate what needs to be done in future.

jenkinsbuilder/cccp_index_reader.py
bamachrn: yes got it, I was asking about putting read old projects and delete projects also to keep in try catch. @navidshaikh how do you think? IMO, if some thing happens wrong some where we need to have logs to track that.

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -118,23 +203,41 @@ def main(indexdlocation):
...
-                                    stdout=subprocess.PIPE)
-            proc.communicate()
+            run_command(myargs)
 
+            new_projects_names.append(
+                str(project[0]["project"]["appid"]) + "-" +
+                str(project[0]["project"]["jobid"]) + "-" +
+                str(project[0]["project"]["desired_tag"])
+            )
         finally:
```
dharmit: There's no `except` clause and all `finally:` is doing is `pass`. We might as well get rid of `finally` in that case.

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -118,23 +203,41 @@ def main(indexdlocation):
...
+    # get list of old projects
+    old_projects = get_old_project_list()
+
+    # find stale projects
+    stale_projects = find_stale_projects(
+        old_projects,
+        list(set(new_projects_names))
+    )
+
+    print stale_projects
```
dharmit: Is this a debug statement you forgot to remove?

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -100,16 +108,93 @@ def get_projects_from_index(indexdlocation):
...
+            fin.write(projects_names)
+    except IOError as e:
+        print "Failed to export project names to file %s" % projects_list
+        print "I/O Error {0}:{1}".format(e.errno, e.strerror)
+    except:
+        print "Unexpected error:", sys.exc_info()[0]
+    else:
+        # change the mod of file so that jenkins user can edit it
+        print "Changing file permission 0777 of file %s" % projects_list
+        run_command(["chmod", "0777", projects_list])
```
dharmit: Is there no other alternative to changing the permissions to `0777`? We're giving global read, write, execute permissions which is not a good idea when this moves to prod.

jenkinsbuilder/cccp_index_reader.py
navidshaikh: The run_command will return `out, error`, I am updating PR to print out errors if any.

jenkinsbuilder/cccp_index_reader.py
navidshaikh: I confirmed with @bamachrn that this is taken care in some other PR. So I will remove it.

bamachrn:
#dotests

bamachrn:
#dotests-debug

PR #314, opened 2017-07-20
Hot-fix based on upstream atomic patch

dharmit:
This hot-fix is based on
https://github.com/projectatomic/atomic/pull/1037 and needs to be
removed after atomic package in CentOS gets updated.

### DIFF
--- a/provisions/roles/scanner/tasks/main.yml
+++ b/provisions/roles/scanner/tasks/main.yml
@@ -19,6 +19,13 @@
   sudo: yes
   tags: scanner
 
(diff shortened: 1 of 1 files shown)


bamachrn:
LGTM if the indentation persists

dharmit:
>  if the indentation persists

That's the reason I've added 12 spaces in there. It will work. :wink:

dharmit:
#dotests

bamachrn:
I am merging this, lets mark this for removal after atomic update

PR #317, opened 2017-07-25
Configures email subject prefix

navidshaikh:
This PR improves the option to configure email subject prefix using variable in hosts file.
Earlier there were only two possible options for this, test and production environment.
With this changeset, a user will be able to configure, either production, pre-prod, or test environment and accordingly email can be distinguished.

### DIFF
--- a/ci/lib.py
+++ b/ci/lib.py
@@ -164,6 +164,7 @@ setup_nfs=True
 production=False
 log_level=DEBUG
 openshift_server_ip={openshift_host}
+deployment=ci
 
 [jenkins_master:vars]
 jenkins_private_key_file = jenkins.key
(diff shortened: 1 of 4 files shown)


navidshaikh:
#dotest-debug

navidshaikh:
Changed the variable name from `environment`--to-->`deployment`, since during provisioning ansible was complaining about conflict of environment variable name chosen.

bamachrn:
#dotests

bamachrn:
LGTM 👍

navidshaikh:
#dotests-debug

navidshaikh:
#dotests-debug

bamachrn:
@navidshaikh PR is failing with syntax `the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'deployment' is undefined` . do you want check if variable is getting assigned in CI?

navidshaikh:
@bamachrn : thanks for pointing it out.

>  do you want check if variable is getting assigned in CI?

I need to add entry in `hosts` file content written over [here](https://github.com/CentOS/container-pipeline-service/blob/master/ci/lib.py#L167).

navidshaikh:
This is ready to merge.

PR #320, opened 2017-08-02
Ansible cleanup on nodes for re-provisioning

navidshaikh:
This PR cleans up nodes for re-provisioning.


provisions/roles/cleanup/openshift/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Remove all containers
+  sudo: yes
+  shell: docker rm -f `docker ps -a -q`
```
rtnpro: You should not remove all the containers, but only the openshift containers.

provisions/roles/cleanup/openshift/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  ignore_errors: yes
+
+- name: Remove all images
+  sudo: yes
+  shell: docker rmi -f `docker images -a -q`
```
rtnpro: Removing all images is also unnecessary. We are not cleaning up docker here, but only openshift. If you have some other reason for doing so, I will be eager to listen.

provisions/roles/cleanup/scanner/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Remove all containers
+  sudo: yes
+  shell: docker rm -f `docker ps -a -q`
```
rtnpro: Kill only the scanner containers.

provisions/roles/cleanup/scanner/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+
+- name: Remove all images
+  sudo: yes
+  shell: docker rmi -f `docker images -a -q`
```
rtnpro: Remove only related images.

provisions/roles/cleanup/jenkins/slave/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Remove all containers
+  sudo: yes
+  shell: docker rm -f `docker ps -a -q`
```
rtnpro: Don't kill containers randomly.

navidshaikh:
#dotests-debug

provisions/roles/cleanup/jenkins/master/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Stop jenkins systemd service
+  sudo: yes
+  service: name=jenkins state=stopped
+
+- name: Remove contents under direcotry /var/lib/jenkins/
+  sudo: yes
+  shell: rm -rf /var/lib/jenkins/*
```
bamachrn: IMO, we should not delete /var/lib/jenkins. this will reset all the settings and may be all the jobs will be rebuilt even if there is no change to them.

provisions/roles/cleanup/jenkins/slave/tasks/main.yml
bamachrn: removing containers would not matter much, because provision will bring all the containers back. Though we have to find a to validate the current  jobs is processed before stopping container.

provisions/roles/cleanup/openshift/tasks/main.yml
bamachrn: we should remove all the containers other than openshift ones. but if we re resetting all the environment, we can clear out everything.

provisions/roles/cleanup/openshift/tasks/main.yml
bamachrn: yeah, if we remove all the images other than the openshift ones, that would suffice. @rtnpro we are cleaning up all the images to make sure we are not keeping the images created while building images. Ideally, there should not be any if pods are deleted properly while building. we can also plan for deleting the project to make sure this is taken care it self. but for now, lets remove all the images (or images except openshift ones if we can) as provision will bring everything back.

provisions/roles/cleanup/scanner/tasks/main.yml
bamachrn: @rtnpro here scanners stop by themselves, but sometimes scanner keeps containers even though scan is over. may be we can solve this in the other way round, by making sure scanner is removing all the containers and images once scan is over for some image.

provisions/roles/cleanup/scanner/tasks/main.yml
bamachrn: IMO, this is not required as scanner images gets updated each deployment time. and there is no extra images there(if scanner deletes everything by itself).

provisions/roles/cleanup/jenkins/master/tasks/main.yml
navidshaikh: @bamachrn : Okay. But then if there are code changes related to jenkins and scripts which reside in `/var/lib/jenkins/*` directory, the deployment dev needs to manually clear it, since it is kept as cache and just a normal deployment does not update scripts inside /var/lib/jenkins directory.

provisions/roles/cleanup/scanner/tasks/main.yml
navidshaikh: @bamachrn : sure!

provisions/roles/cleanup/openshift/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    name: mail-server
+    state: absent
+    force_kill: yes
+
+- name: Stop and remove OpenShift origin container
+  sudo: yes
+  docker_contianer:
+    name: origin
+    state: absent
+    force_kill: yes
```
bamachrn: We have to add registry and router also here and pod deployer will also not be in exited state. please `docker rm -f $(docker ps -qa)` would help here. but I would suggest to have a setup running somewhere and write this, rather than putting thoughts on this

bamachrn:
LGTM 👍

PR #323, opened 2017-08-07
Updates ansible docker module operations

navidshaikh:
Ansible `docker` module is deprecated, `docker_image` and `docker_container` are the new ansible-docker operation module names. Ported old ansible tasks from `docker` and shell scripts to `docker_image` and `docker_container`.

### DIFF
--- a/provisions/roles/jenkins/slave/tasks/setup_worker.yml
+++ b/provisions/roles/jenkins/slave/tasks/setup_worker.yml
@@ -19,48 +19,49 @@
   file: dest=/opt/cccp-service/beanstalk_worker state=directory
 
 - name: Build container-pipeline image
-  shell: docker build -t container-pipeline -f ./Dockerfile .
-  args:
-      chdir: /opt/cccp-service
+  sudo: yes
+  docker_image:
+    name: container-pipeline
+    path: /opt/cccp-service
   tags:
       - application
 
 - name: Ensure log path exists
   file: path=/srv/pipeline-logs/cccp.log state=touch
 
-- name: Start dispatcher worker
-  docker:
+- name: Start dispatcher-worker container
+  docker_container:
       name: dispatcher-worker
-      state: restarted
+      state: present
+      restart: yes
       image: container-pipeline
       command: /opt/cccp-service/container_pipeline/workers/dispatcher.py
       restart_policy: always
-      volumes:
-        /srv/pipeline-logs:/srv/pipeline-logs:rw
+      volumes: /srv/pipeline-logs:/srv/pipeline-logs:rw
   tags:
       - application
 
-- name: Start build worker
-  docker:
+- name: Start build-worker container
+  docker_container:
       name: build-worker
-      state: restarted
+      state: present
+      restart: yes
       image: container-pipeline
       command: /opt/cccp-service/container_pipeline/workers/build.py
       restart_policy: always
-      volumes:
-        /srv/pipeline-logs:/srv/pipeline-logs:rw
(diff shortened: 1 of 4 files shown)


navidshaikh:
#dotests-debug

navidshaikh:
@dharmit @bamachrn @mohammedzee1000 PTAL

provisions/roles/jenkins/slave/tasks/setup_worker.yml
```diff
@@ -19,48 +19,49 @@
   file: dest=/opt/cccp-service/beanstalk_worker state=directory
 
 - name: Build container-pipeline image
-  shell: docker build -t container-pipeline -f ./Dockerfile .
-  args:
-      chdir: /opt/cccp-service
+  sudo: yes
```
dharmit: Provisioning's happening as `root` user, right? Why use `sudo` then?

provisions/roles/openshift/tasks/setup_mailservice.yml
```diff
@@ -3,27 +3,26 @@
...
       dest="{{ ansible_env.HOME }}/cccp-service/mail_service/config.py"
       regexp="LOG_LEVEL =.*"
       replace="LOG_LEVEL = '{{ log_level }}'"
 
-- name: Build mail server image
-  shell: docker build -t mail-server -f ./Dockerfile.mailserv .
-  args:
-      chdir: "{{ ansible_env.HOME }}/cccp-service/mail_service"
+- name: Build mail-server container image
+  sudo: yes
```
dharmit: Same comment as provisioning as `root` user.

provisions/roles/jenkins/slave/tasks/setup_worker.yml
navidshaikh: this wont have effect if user is `root`, but it enables our service to be deployed on environments where sudo access is not given to devs.

dharmit: Do you mean that if a dev is allowed to login to system X as user `dev` and if the user `dev` doesn't have `sudo` privileges on that system, the `sudo: yes` directive will help bypass that and start the container?

navidshaikh: Not really!

this is for user having only `sudo` access and provisioning happening with non-root account.

dharmit:
LGTM

provisions/roles/openshift/tasks/config.yml
```diff
@@ -33,17 +33,17 @@
...
     privileged: yes
-    pid: host
-    net: host
+    pid_mode: host
+    network_mode: host
     volumes: "{{ openshift_volumes }}"
     command: start --master=https://"{{ansible_default_ipv4.address}}":8443
     restart_policy: always
-  retries: 2
+    restart_retries: 2
```
bamachrn: @navidshaikh may be we do not change anything for openshift here, this is being taken care in the openshift upgrade PR also. there are few more changes. this will cause merge conflicts. also IMO, we can put all the openshift related changes to single PR.

navidshaikh: @bamachrn 
This is not a change, the `retries` keyword is renamed to `restart_retries` wiht newer `docker_container` keyword. This PR just migrating older keywords to newer ones, and whatever change latest openshift PR would do, will be on top of it. I dont think two PRs should be corelated.

bamachrn: @navidshaikh my comment was for the whole block. any way that makes sense. we can merge it in may be. thanks 👍

PR #329, opened 2017-08-17
Fix worker to take care the job flow

bamachrn:
* This will make the job flow from jenkins to notification tube.
* delivery worker is sending notifications 
* shell script interaction with jobs is removed


container_pipeline/workers/delivery.py
```diff
@@ -55,7 +56,8 @@ def deliver_build(self, job):
             if not delivery_id:
                 return False
         finally:
-            self.openshift.delete_pods(project, delivery_id)
+            time.sleep(50)
+            self.openshift.delete(project)
```
dharmit: Can we add a note (comment) about why we're deleting an OpenShift project?

container_pipeline/workers/delivery.py
```diff
@@ -55,7 +56,8 @@ def deliver_build(self, job):
             if not delivery_id:
                 return False
         finally:
-            self.openshift.delete_pods(project, delivery_id)
+            time.sleep(50)
```
navidshaikh: Why are we adding this 50 seconds delay ? because openshift will refuse to delete it immediately ? (please add comment there as well)

navidshaikh:
@bamachrn : tests!

PR #330, opened 2017-08-18
Updated the build process to update openshift source before build

mohammedzee1000:
This is for the dealing with dockerfiles not requiring tags in FROM stanza.

### DIFF
--- a/server/run.sh
+++ b/server/run.sh
@@ -56,8 +56,10 @@ _ "Put build,test, delivery scripts in proper place"
 # Alter target file, if openshift container is detected
 CONTAINER_NAME="${APPID}/${JOBID}";
 if [[ ${CONTAINER_NAME} =~ ^openshift\/(origin.*|node|openvswitch) ]]; then
-    _ "Openshift container detected, altering Target File ${TARGET_FILE}";
-    sed -i "/^FROM/ s/$/:${DESIRED_TAG}/" ${TARGET_FILE};
+    _ "Openshift container detected, checking if  ${TARGET_FILE} needs to be altered";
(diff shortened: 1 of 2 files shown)


server/run.sh
```diff
@@ -53,6 +53,12 @@ _ "Put build,test, delivery scripts in proper place"
 #echo "ADD /test_script /usr/bin/" >> $TARGET_FILE
 #echo "ADD /delivery_script /usr/bin/" >> $TARGET_FILE
 
+# Alter target file, if openshift container is detected
+CONTAINER_NAME="${APPID}/${JOBID}";
```
navidshaikh: +1 on checking for matching the resulting container name

bamachrn:
LGTM 👍

bamachrn:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests-debug

server/run.sh
```diff
@@ -53,6 +53,15 @@ _ "Put build,test, delivery scripts in proper place"
...
+if [[ ${CONTAINER_NAME} =~ ^openshift\/(origin.*|node|openvswitch) ]]; then
+    _ "Openshift container detected, checking if  ${TARGET_FILE} needs to be altered";
+    if [[ ( ${DESIRED_TAG} == "v1.5" && ${CONTAINER_NAME} != "openshift/origin-base" ) && ${CONTAINER_NAME} != "openshift/origin-source" ]]; then
```
bamachrn: why are we hardcoding to 1.5 here. cant we have any <desired_tag>. and exception for `latest` only?

mohammedzee1000: @bamachrn we are only doing 1.5 and above and only 1.5 has the origin-base coming from centos 3.6 onwards, its origin-source

PR #332, opened 2017-08-21
Fix issues with RPM tracking feature

dharmit:
RPM tracking feature had following issues:

- `null` values for `mirrorlist` were getting inserted into the database
- `Index out of range` issue when comparing old and new packages
- `RepoMDError` when the repo being checked was that of EPEL

### DIFF
--- a/container_pipeline/management/commands/imagescanner.py
+++ b/container_pipeline/management/commands/imagescanner.py
@@ -42,6 +42,10 @@ def populate_upstreams(container):
         urls.add(item[1])
         for url in item[2]:
             urls.add(url)
+
+    if None in urls:
+        urls.remove(None)
+
     output = container.run(
         'python -c "import yum, json; yb = yum.YumBase(); '
         'print json.dumps(yb.conf.yumvar)"')
--- a/container_pipeline/management/commands/imagescanner.py
+++ b/container_pipeline/management/commands/imagescanner.py
@@ -34,7 +34,9 @@ def populate_upstreams(container):
     logger.debug('Populating upstream data for image: %s' % container)
     output = container.run(
         'python -c "import yum, json; yb = yum.YumBase(); '
-        'print json.dumps([(r.id, r.mirrorlist, r.baseurl) '
+        'print json.dumps([(r.id,'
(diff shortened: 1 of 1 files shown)


dharmit:
I have opened a single PR to fix multiple issue because fixing one issue brought up another and it wasn't until all things were taken care of that we started seeing error-free execution of RPM tracking feature.

container_pipeline/management/commands/imagescanner.py
```diff
@@ -34,14 +34,20 @@ def populate_upstreams(container):
     logger.debug('Populating upstream data for image: %s' % container)
     output = container.run(
         'python -c "import yum, json; yb = yum.YumBase(); '
-        'print json.dumps([(r.id, r.mirrorlist, r.baseurl) '
+        'print json.dumps([(r.id,'
+        '\'https://mirrors.fedoraproject.org/mirrorlist?repo=epel-7&arch=x86_64\','
+        'r.baseurl) if r.id==\'epel\' else (r.id, r.mirrorlist, r.baseurl) '
         'for r in yb.repos.listEnabled()])"')
```
bamachrn: for now we are putting this for `epel` only. do we want to make a dictionary for these exclusions later?

container_pipeline/management/commands/imagescanner.py
```diff
@@ -34,14 +34,20 @@ def populate_upstreams(container):
...
     data = json.loads(output.splitlines()[-1])
     urls = set()
     for item in data:
         urls.add(item[1])
         for url in item[2]:
             urls.add(url)
+
+    if None in urls:
```
bamachrn: this is not coming from DB right? or we should consider avoiding `NULL` also?

container_pipeline/management/commands/imagescanner.py
dharmit: We can do that later if similar issue pops up for any other repo.

container_pipeline/management/commands/imagescanner.py
dharmit: This check is made before we commit anything to the DB. So a null value should never go into the DB.

PR #334, opened 2017-08-22
Copies beanstalkc for weekly scan in during provisioning

navidshaikh:
This change copies beanstalkc.py in jenkinsbuilder directory
  during provisioning. This was a bug and needed fix for weekly scan
  to run. weekly scan script was erroring out failing to import beanstalkc


bamachrn:
LGTM

PR #335, opened 2017-08-23
Populate DB with image names before triggering build

dharmit:
This is to ensure that when image_scanner worker checks for image name
in the DB, it's actually there.

### DIFF
--- a/ci/tests/test_01_functional/test_01_repo_monitoring/__init__.py
+++ b/ci/tests/test_01_functional/test_01_repo_monitoring/__init__.py
@@ -35,28 +35,6 @@ class TestRepoMonitoring(BaseTestCase):
             'cd /opt/cccp-service && '
             'python -c "{}"'.format(_script))
 
-    def test_00_if_fetch_scan_image_job_is_successful(self):
-        self.run_cmd(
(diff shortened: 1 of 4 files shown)


dharmit:
@rtnpro need your review on this.

bamachrn:
LGTM 👍

rtnpro:
Why do we need to do it during every build? AFAIK, the ``fetchimagelist`` is called post container-index job as a post build hook. This ensures that we always have an updated list of images in db.

dharmit:
> Why do we need to do it during every build? AFAIK, the fetchimagelist is called post container-index job as a post build hook. This ensures that we always have an updated list of images in db.

So the issue is that `fetch-scan-image` job does two things:

- populate the db
- do one time check of the image through imagescanner.

2nd part of it is not required because the image_scanner on jenkins master already does that post delivery phase. Also, when we run one time check right after populating the db, we run into an issue where if the image is not already "built, tested, scanned and delivered", it doesn't appear on the registry and the one time scan fails.

So what I've done is, move the populating DB part of it to `cccp-index` job and remove the onetime scan part completely as we already do that post delivery.

PR #337, opened 2017-08-24
Removed OpenShift project deletion while it is building

bamachrn:
Delivery build is deleting the project from openshift once the build is over.  So if the project exists in openshift we should wait for the previous job to get over, otherwise job parameters will get replaced.
And then also if it is not getting deleted that means there is some issue in openshift. So we log it for developers and try with next job.


bamachrn:
#dotests

container_pipeline/trigger_build.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import time
+
+import container_pipeline.utils as utils
+from container_pipeline.lib.openshift import Openshift, OpenshiftError
+
+
+def create_project(queue, job, logger):
+    job_name = job.get("job_name")
+    project = utils.get_job_hash(job_name)
```
dharmit: We're storing hash value of the job in the variable `project`. This makes things a bit ambiguous. Can we use a more relevant variable name instead?

container_pipeline/trigger_build.py
```diff
@@ -0,0 +1,66 @@
...
+        })
+    except OpenshiftError:
+        try:
+            openshift.delete(project)
+        except OpenshiftError as e:
+            logger.error(e)
+        return
+
+    job["action"] = "start_build"
+    logger.info("Putting job to build tube: {}".format(job))
```
dharmit: We're putting job to the `master_tube`. This would mislead dev/admin during troubleshooting/debugging.

provisions/roles/cleanup/jenkins/slave/tasks/main.yml
```diff
@@ -1,21 +1,28 @@
 ---
```
dharmit: Using `with_items`, we can make this a single task. Is there a reason we're doing this as separate jobs for each container?

bamachrn: that was extension of the previous cleanup code. fixing this

PR #338, opened 2017-08-28
[WIP] Synchronize key-value pairs across workers

dharmit:


container_pipeline/pipeline.py
```diff
@@ -3,60 +3,137 @@
...
-    job_name = utils.get_job_name({
-        'appid': appid, 'jobid': jobid, 'desired_tag': desired_tag})
-    project = utils.get_job_hash(job_name)
+def create_new_job():
+    """
+    Creates new job (a dictionary) with default value as `None` for all the
+    keys. This is the central place for creation of a job for container
+    pipeline service
+    """
+    job = dict.fromkeys([
```
bamachrn: grouping the similar/related keys would be better.

container_pipeline/workers/build.py
```diff
@@ -27,14 +27,16 @@ def handle_job(self, job):
         queue to be processed later. Else, it goes ahead with running
         build for the job.
         """
+        self.job = job
```
bamachrn: this wont break the inheritance right? I mean base worker calls the handle build job from run method.

container_pipeline/workers/test.py
```diff
@@ -45,56 +45,58 @@ def run_test(self, job):
...
+        self.job['output_image'] = \
+            "registry.centos.org/{}/{}:{}".format(
+                self.job['appid'], self.job['jobid'], self.job['desired_tag'])
+        self.job['beanstalk_server'] = settings.BEANSTALKD_HOST
+        self.job['image_name'] = \
+            "{}/{}:{}".format(
+                self.job['appid'], self.job['jobid'], self.job['desired_tag'])
+
+        self.queue.put(json.dumps(self.job), 'master_tube')
         self.logger.debug("Test is successful going for next job")
```
bamachrn: if we can set all this in linter worker or in the job initialization only. IMO, setting value for parameter in the worker should be there only that worker is changing the value. and these are no change

container_pipeline/pipeline.py
dharmit: how about making it chronological to make searching/accessing easier? Like this:
```python
job = dict.fromkeys(['action',
             'appid',
             'beanstalk_server',
             'build_status',
             'delivery_status',
             'depends_on',
             'desired_tag',
             'dockerfile',
             'image_name',
             'image_under_test',
             'job_name',
             'jobid',
             'last_run_timestamp',
             'lint_status',
             'logs_URL',
             'logs_dir',
             'logs_file_path',
             'namespace',
             'notify_email',
             'output_image',
             'project_hash_key',
             'project_name',
             'repo_branch',
             'repo_build_path',
             'repo_url',
             'retry',
             'retry_delay',
             'scan_status',
             'target_file',
             'test_tag'])
```

container_pipeline/workers/build.py
dharmit: No this won't break inheritance. Base worker calls `handle_job` function with a parameter called `job` while we're assigning its value to a instance variable which has value `None`. In fact, I'm planning to move `self.job = None` from workers' `__init__` method to the base class's `__init__` method.

container_pipeline/workers/test.py
dharmit: That's the anomaly I was talking about. Clearly I missed it. Thank you!

dharmit:
Closing this as I just needed preliminary feedback. Will open a new PR once the task has been accomplished for all workers.

PR #339, opened 2017-08-31
Synchronize key-value pairs across workers

dharmit:


bamachrn:
#dotests

bamachrn:
#dotests-debug

bamachrn:
#dotests-debug

bamachrn:
#dotests-debug

bamachrn:
#dotests-debug

container_pipeline/trigger_dockerfile_lint.py
```diff
@@ -81,20 +48,20 @@ def trigger_dockerfile_linter(appid, jobid, repo_url, repo_branch,
         print "==> Sending Dockerfile linter failure email"
         response = {
```
bamachrn: We should remove this job creation and use the job variable changing the action and setting msg. May be we can add `msg` to dictionary also. We can take care of these while re factoring the mail service as well. (All the separate job creation left in this PR, are related to `notify_user`)

dharmit: As discussed, let's keep this in a separate PR wherein we refactor the `notify_user` piece.

PR #340, opened 2017-09-08
Add information to status email about cause of build

dharmit:
Before:
```
Subject: SUCCESS: Container build: devcloud/wget:latest is complete                                                             
                                                                                                                                                                             
CentOS Community Container Pipeline Service <https://github.com/centos/container-index>                                                                                      
=================================================================                                                                                                            
                                                                                                                                                                             
Build status:           Success                                                                                                                                              
Image:                  registry.centos.org/devcloud/wget:latest                                                                                                             
Build logs:             https://registry.centos.org/pipeline-logs/NzNkNjA1MTkxNW
```

After:
```
Subject: SUCCESS: Container build: devcloud/wget:latest is complete                                                       
                                                                                                                                                                             
CentOS Community Container Pipeline Service <https://github.com/centos/container-index>                                                                                      
=================================================================                                                                                                            
                                                                                                                                                                             
Build status:           Success                                                                                                                                              
Image:                  registry.centos.org/devcloud/wget:latest                                                                                                             
Build logs:             https://registry.centos.org/pipeline-logs/NzM2ZGE2NzRlMT                                                                                             
Cause of build:         Git commit 57873245041f35cea257672a40bd11f0330cb96d
```

### DIFF
--- a/container_pipeline/utils.py
+++ b/container_pipeline/utils.py
@@ -1,6 +1,8 @@
 import hashlib
+import json
 import logging
 import os
+import urllib2
 
 
 def get_job_name(job_details):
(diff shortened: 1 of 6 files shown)


container_pipeline/utils.py
```diff
@@ -20,6 +22,45 @@ def get_project_name(job):
...
+    the cause of build
+    """
+    try:
+        url = "http://{}:8080/job/{}/{}/api/json".format(
+            jenkins_url, job_name, job_number)
+        response = urllib2.urlopen(url)
+    except URLError:
+        return "Invalid URL: {}".format(url)
+    else:
+        return parse_json_response(json.loads(response.read()))
```
bamachrn: may be we can keep `parse_json_respnse ` inside of try. so that we can the parse error also if there is any.

container_pipeline/utils.py
```diff
@@ -20,6 +22,45 @@ def get_project_name(job):
     return '{}-{}-{}'.format(job['appid'], job['jobid'], job['desired_tag'])
 
 
+def parse_json_response(response):
+    """Parses the json response provided to it to determin"""
+    try:
+        cause = response['actions'][0]['causes'][0]['shortDescription']
+
+        if cause == 'Started by an SCM change':
+            return 'Git commit {}'.format(
```
bamachrn: do we want to keep the line as "Triggered by SCM change with GIT commit id: <commit id>"?

dharmit: That's what I had initially done. But in the mail, that would then show up as something like:

`Cause of build:  Triggered by SCM change with GIT commit id: abcdef`.

`Cause of build` and `Triggered by` indicate the same thing. So I went with mentioning only `Git commit` approach.

container_pipeline/utils.py
dharmit: We have a `try..except` block in `parse_json_response` to handle its only possible `KeyError`. So I kept that in `else` block.

PR #341, opened 2017-09-14
Added model for storing information about the container.

mohammedzee1000:
This is for the purpose of storing a data about the container, so that the UI can access the same and allow the user to view the source files of the containers they are downloading.


navidshaikh:
#dotests-debug

provisions/roles/jenkins/common/tasks/main.yml
```diff
@@ -8,6 +8,16 @@
...
 
+    #- name: Ensure pip is installed.
+    #  yum: name=python-pip state=present
+    #  tags:
+    #    - jenkins
+    #    - jenkins/master
+    #    - jenkins/slave
+
+      #- name: Ensure django is installed
+      #  shell: "pip install django"
```
dharmit: Are we sure that not installing django on jenkins master & slave is OK?

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,83 @@
...
+    """Represents information about the container, stored in the model."""
+
+    def __init__(self, data_dir="/srv/pipeline-logs"):
+        """
+        Intialize the Model
+        :param data_dir: The directory to which we need to dump the data file to.
+        """
+        self._data_dir = data_dir
+        self.data_file = data_dir + "/container_info.json"
+        self.data = {
```
dharmit: Safer way would be to do `self.data = dict()`.

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,83 @@
...
+    :param git_path: The path, relative to the git repository root, where the file resides.
+    :param git_branch: The repository branch where the file resides.
+    :param target_file: The name of the target file.
+    :return: The reachable link to the dockerfile.
+    """
+
+    # TODO : Move this into a lib.
+
+    link_url = None
+    if "github" in git_url or "gitlab" in git_url:
```
dharmit: do we really want to exclude all hosted git platforms except github and gitlab? cc @bamachrn

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,83 @@
...
+            dump(self.data, f)
+
+    def append_info(self, container_name, dockerfile_link):
+        """
+        Appends the container information to the model
+        :param container_name: The name of the container
+        :param dockerfile_link: The link to the dockerfile.
+        """
+        if container_name not in self.data:
+            self.data[container_name] = {
```
dharmit: Do we want to create a key with the value of parameter `container_name` or with the value `"container_name"`?

provisions/roles/jenkins/common/tasks/main.yml
mohammedzee1000: I actually found a task that installs django on one node some time back, so thanks i will remove this and update the task for more nodes.

container_pipeline/model_tmp/containers.py
mohammedzee1000: @dharmit for now. yes. I will need to go through all the available git repos and figure out that their git path will look like. Thanks for pointing it out though i will put in an else condition

container_pipeline/model_tmp/containers.py
mohammedzee1000: Hmm, yea i see that repetition is there. What do you think of using a hash? or maybe removing the field oor container_name and just having key tell me that?

The dict would make more search to reduce search time on UI side.

container_pipeline/model_tmp/containers.py
mohammedzee1000: +1 will update accordingly

container_pipeline/model_tmp/containers.py
dharmit: I don't understand the reasoning about going through all available git repos. You plan to go through all available git repos for the projects that have their entries on container-index or, do you mean something else?

container_pipeline/model_tmp/containers.py
dharmit: I think my concern was misunderstood. For example, I'd expect value of variable `container_name` to be something like `dharmit/python:3.5`. My question is, are you sure you want to have `self.data["dharmit/python:3.5"]` and not `self.data["container_name"]`? 

This is mainly for my fear wherein we merge something in the master that might become non-trivial to debug.

container_pipeline/model_tmp/containers.py
mohammedzee1000: No, i mean going through all the entries in the index, irrespective of the git repo they come from and having the necessary information updated in the model.

For now, we have only github and gitlab urls, but i need to go through other hosted git solutions, and figure out what their git paths will look like, so link for the same can be formed.

container_pipeline/model_tmp/containers.py
mohammedzee1000: Ah, i see where you are going. Will update.

navidshaikh:
@mohammedzee1000 : is this ready to merge ?

container_pipeline/model_tmp/containers.py
bamachrn: @mohammedzee1000 @dharmit I agree from the feasibility point of view for now. but did we look on other SCM? I believe we can generalize this as required.

container_pipeline/model_tmp/containers.py
bamachrn: Also we need match this for the primary key in the DB which @navidshaikh is working on. this would help putting proper things/ links in right contianer.

mohammedzee1000: I believe the full container name would be the most appropriate primary key, as in "appid/jobid:desiredtag". We can always split if we need the appid, jobid and desiredtag seperately

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+#from django.conf import settings
+from os import path
+from json import load, dump
+
+# TODO : Change implementation into proper model and move this class into model package.
+
+
+def form_dockerfile_link(git_url, git_path, git_branch, target_file):
+    """
+    Helper function to generate docker file link.
```
navidshaikh: *Dockerfile

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+    :param git_url: The url of the git repository.
```
navidshaikh: *URL

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+            git_path=git_path,
+            target_file=target_file
+        )
+    return link_url
+
+
+class ContainerLinksModel(object):
+    """Represents information about the container, stored in the model."""
+
+    def __init__(self, data_dir="/srv/pipeline-logs"):
```
navidshaikh: define this at the top of (this) file like
```
DATA_FILE_DIR_DEFAULT = "/srv/pipeline-logs"
```
and then use the constant (written in caps) as the default value for parameter of the function.

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+        self._data_dir = data_dir
+        self.data_file = data_dir + "/container_info.json"
+        self.data = dict()
+        if path.exists(self.data_file):
+            self.unmarshall()
+        else:
+            self.marshall()
+
+    def unmarshall(self):
+        """Loads the modal information into the object."""
```
navidshaikh: typo!
*model

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+
+        with open(self.data_file, "r") as f:
```
navidshaikh: unnecessary declaring mode ```"r"``` , default is read only.

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+            self.data = load(f)
```
navidshaikh: I would suggest you  ```import json``` and then while loading write, ```json.load``` , for the sake of readability (same applies to ```dump``` as well.

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def marshall(self):
+        """Saves the model information"""
+
+        with open(self.data_file, "w+") as f:
```
navidshaikh: only ```w``` is needed.

container_pipeline/model_tmp/containers.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        Appends the container information to the model
+        :param container_name: The name of the container
+        :param dockerfile_link: The link to the dockerfile.
+        """
+        if container_name not in self.data:
+            self.data[container_name] = {
+                "dockerfile_link": ""
+            }
+        self.data[container_name]["dockerfile_link"] = dockerfile_link
```
navidshaikh:
```
sef.data[container_name] = {"dockerfile_link": dockerfile_link}
```
is easier.

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -114,6 +120,15 @@ def get_projects_from_index(indexdlocation):
                         if dependson_job == '':
                             dependson_job = 'none'
 
+                        container_name = str.format(
+                            "{app_id}/{job_id}:{desired_tag}",
+                            app_id=appid,
+                            job_id=jobid,
+                            desired_tag=desiredtag
```
navidshaikh: this string concatenation can be simpler.

mohammedzee1000:
@navidshaikh updated pr as per your comments.

mohammedzee1000:
#dotests

PR #343, opened 2017-09-18
deployment: systemd services reload and clean up tweaks

navidshaikh:
This PR aims to improve the service deployment via ansible.

  - Reloads all the cccp-service systemd services before
  restarting them during deployment.

  - Improves clean up module to remove stale code while ran clean up module before deployment

### DIFF
--- a/provisions/roles/cleanup/jenkins/master/tasks/main.yml
+++ b/provisions/roles/cleanup/jenkins/master/tasks/main.yml
@@ -10,4 +10,4 @@
 - name: Remove images with no name or tag
   sudo: yes
   ignore_errors: yes
-  shell: "docker rmi -f $(docker images | awk '/<none>/ {print $3}')"
+  shell: docker rmi $(docker images -f "dangling=true" -q)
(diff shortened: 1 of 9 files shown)


navidshaikh:
#dotests

provisions/roles/cleanup/jenkins/master/tasks/main.yml
```diff
@@ -2,3 +2,12 @@
...
   service: name=jenkins state=stopped
+
+- name: Remove the virtual environment directories
+  sudo: yes
+  file: path=/opt/cccp-service state=absent
+
+- name: Remove images with no name or tag
+  sudo: yes
+  ignore_errors: yes
+  shell: "docker rmi -f $(docker images | awk '/<none>/ {print $3}')"
```
dharmit: IMHO, it's better to do `docker rmi $(docker images -q -f dangling=true)` and let Docker daemon decide which images are safe to delete.

provisions/roles/jenkins/master/tasks/install.yml
```diff
@@ -15,5 +15,9 @@
 - name: Ensure Jenkins is installed.
   yum: name=jenkins-2.67-1.1 state=present
 
+- name: Reload jenkins systemd service
+  sudo: yes
+  service: name=jenkins state=reloaded
```
dharmit: Based on https://github.com/ansible/ansible-modules-core/issues/3897#issuecomment-224893624, I'd recommend we use `systemd` module instead of `service` module.

dharmit:
#dotests

PR #344, opened 2017-09-20
Add registry config if not found

bamachrn:


provisions/roles/openshift/tasks/config.yml
```diff
@@ -1,15 +1,17 @@
 ---
 - name: Enable Docker insecure registry
-  replace: >
-    dest=/etc/sysconfig/docker
+  lineinfile:
+    path=/etc/sysconfig/docker
+    state=present
     regexp="^#*\s*INSECURE_REGISTRY=.*"
```
dharmit: With the latest version of Docker available on CentOS 7, I saw that this line doesn't exist at all. Not even commented, as was the case earlier. So, I had to leave the `regexp` part altogether. Here's what I did: https://github.com/dharmit/digitalocean-compass-ansible/blob/master/main.yml#L61-L64

dharmit: So with `state=present`, I think, the `regexp` you've mentioned won't be found and hence the `line` you want to add to the file won't be added.

provisions/roles/openshift/tasks/config.yml
```diff
@@ -1,15 +1,17 @@
...
-    replace='INSECURE_REGISTRY="--insecure-registry 172.30.0.0/16 --insecure-registry {{ intranet_registry }}"'
+    line='INSECURE_REGISTRY="--insecure-registry 172.30.0.0/16 --insecure-registry {{ intranet_registry }}"'
 
 - name: Enable Docker registry
-  replace: >
-    dest=/etc/sysconfig/docker
+  lineinfile:
+    path=/etc/sysconfig/docker
+    state=present
     regexp="^#*\s*ADD_REGISTRY=.*"
```
dharmit: Same comment as for `INSECURE_REGISTRY`. There's nothing like `# ADD_REGISTRY` in the `/etc/sysconfig/docker` file.

provisions/roles/openshift/tasks/config.yml
bamachrn: ok but the ansible documentation says it adds if not found. I know it is not there and that is the reason we are hitting the issue

provisions/roles/openshift/tasks/config.yml
bamachrn: sure updating the same

provisions/roles/openshift/tasks/config.yml
dharmit: English isn't my first language and I admit that the documentation is kinda confusing me as well. From the docs:
> For `state=present`, the pattern to replace if found; only the last line found will be replaced. 

Now the `regext` won't be found as required by `the pattern to replace if found`, and as a result, line we're looking to add won't be added. That's how I interpret the statement. You can see in the link I mentioned in first comment that there's no `state` or `regexp` parameter. But, I could be wrong. :smile:

bamachrn:
#dotests

bamachrn:
#dotests-debug

PR #346, opened 2017-09-28
[WIP]Removed rsync_ssh opts and updated ansile version to 2.3.2

bamachrn:


bamachrn:
NFS setup is failing with updated ansible as I am facing since last few days. I am on it

bamachrn:
closing it as we need to rethink about these

PR #353, opened 2017-10-11
Set valid mirrorlist/baseurl URLs for alternate yum repos

dharmit:
In this PR, I've created a separate dictionary for the URLs of repos that we find not helpful in finding `repomd.xml` file or `repodata` directory. In future, if we encounter more repositories showing similar behavior as EPEL and remi-repos, we can explicitly add their URLs.

Not a perfect solution but, I think there are more than one ways of having a valid `mirrorlist`/`baseurl` in a `.repo` file.


dharmit:
#dotests

dharmit:
#dotests

navidshaikh:
#dotests-debug

PR #354, opened 2017-10-12
Added ability to detect library images to garbage collector.

mohammedzee1000:
This PR adds the ability to detect and clean up library images to garbage collector.

### DIFF
--- a/container_pipeline/cleanup_registry/garbagecollector.py
+++ b/container_pipeline/cleanup_registry/garbagecollector.py
@@ -54,16 +54,16 @@ class GarbageCollector(object):
         registry_storage_path = "/var/lib/registry/docker/registry/v2"
         registry_blobs = registry_storage_path + "/blobs"
         registry_repositories = registry_storage_path + "/repositories"
-        for k, v in self.mismatched.iteritems():
+        for container_full_name, tag_list in self.mismatched.iteritems():
             # For every entry in mismatched
             ## Formulate nessasary data
-            namespace = k.split("/")[0]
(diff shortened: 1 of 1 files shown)


mohammedzee1000:
#dotests

container_pipeline/cleanup_registry/garbagecollector.py
```diff
@@ -54,16 +54,16 @@ def _delete_mismatched(self):
...
         registry_repositories = registry_storage_path + "/repositories"
-        for k, v in self.mismatched.iteritems():
+        for container_full_name, tag_list in self.mismatched.iteritems():
             # For every entry in mismatched
             ## Formulate nessasary data
-            namespace = k.split("/")[0]
+            namespace = container_full_name.split("/")[0] if "/" in container_full_name else container_full_name
             namespace_path = registry_repositories + "/" + namespace
-            container_name = registry_repositories + "/" + k
+            container_name = registry_repositories + "/" + container_full_name
```
bamachrn: @mohammedzee1000 would you please post some examples for the value of `container_name`in case of library and non library containers. that would make this part more clear. as we should cautious while deleting images from registry

mohammedzee1000: @bamachrn 
Ok so the way the garbage collector basically works is go through the index, go through the registry. Find all the images, in registry, but not in index, and remove them. 

When it comes to normal images, a simple match will do. However, in case of library images, the  image on registry will not have an app-id, while it does in the index. 

So basically, in my change. What i have done is basically, while going through index, if the image has app-id of 'library', then it is ignored while noting down name of image, to match with registry.

PR #356, opened 2017-10-21
Pipeline data storage

mohammedzee1000:
This implements saving pipeline state/data to DB so that we can churn various useful metrics from it, and also make our workers fault tolerant.


mohammedzee1000:
ok to test

mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

navidshaikh:
#dotests-debug

navidshaikh:
#dotests-debug

provisions/roles/db/tasks/main.yml
```diff
@@ -12,6 +12,13 @@
   sudo: yes
   tags: db
 
+- name: Reload docker systemd service
+  sudo: yes
+  systemd: name=docker state=reloaded
+
+- name: Start Docker service
```
navidshaikh: add 

sudo: yes

navidshaikh: also should it be restarted or started ?

provisions/roles/repo_tracking/tasks/main.yml
```diff
@@ -53,6 +53,14 @@
   sudo: yes
   tags: repo_tracking
 
+- name: Initialize for creating DB tables.
+  shell: /opt/cccp-service/manage.py makemigrations --noinput
+  ignore_errors: yes
```
navidshaikh: we can afford to ignore tables ?

provisions/roles/repo_tracking/tasks/main.yml
```diff
@@ -53,6 +53,14 @@
...
+  retries: 5
+  delay: 10
+  sudo: yes
+  tags: repo_tracking
```
navidshaikh: is this correct tag ?

provisions/roles/repo_tracking/tasks/main.yml
```diff
@@ -53,6 +53,14 @@
   sudo: yes
   tags: repo_tracking
```
navidshaikh: typos in commit message

container_pipeline/management/commands/triggerbuilds.py
```diff
@@ -6,7 +6,8 @@
 from django.utils import timezone
 from django.conf import settings
 
-from container_pipeline.models.tracking import ContainerImage, Package
```
navidshaikh: this should have just worked too!

container_pipeline/models/pipeline.py
```diff
@@ -83,6 +83,7 @@ class BuildPhase(models.Model):
     STATUS_CHOICES = (
         ('queued', 'Queued'),
         ('processing', 'In process'),
+        ('requeuedparent', 'Requeued due to parent'),
```
navidshaikh: very good!

container_pipeline/management/commands/triggerbuilds.py
mohammedzee1000: True, i just refactored the placement based on specialisation of the classes, so it makes sense programatically. I think ContainerImage can be used to store more than just repo tracking information.

navidshaikh:
#dotests-debug

container_pipeline/workers/linter.py
```diff
@@ -40,13 +41,11 @@ def handle_job(self, job):
...
-            build=build, phase='dockerlint')
-        self.build_phase.status = 'processing'
-        self.build_phase.start_time = timezone.now()
-        self.build_phase.save()
-
+        self.setup_data()
+        self.set_data(
+            build_phase_status='processing',
+            build_phase_start_time=timezone.now()
+        )
```
navidshaikh: good!

container_pipeline/workers/build.py
```diff
@@ -4,10 +4,13 @@
...
 import time
 
+from container_pipeline.lib import dj  # noqa
+from django.utils import timezone
+
 from container_pipeline.lib import settings
 from container_pipeline.lib.log import load_logger
 from container_pipeline.lib.openshift import Openshift, OpenshiftError
-from container_pipeline.utils import Build, get_cause_of_build
+from container_pipeline.utils import Build as BuildTracker, get_cause_of_build
```
navidshaikh: this could bring ambiguity, how about changing the rename of `Build` class / model itself?

mohammedzee1000:
#dotests-debug

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

container_pipeline/models/pipeline.py
```diff
@@ -0,0 +1,209 @@
...
+        db_table = 'builds'
+
+    def __str__(self):
+        return '{}:{}'.format(self.project, self.uuid)
+
+
+class BuildPhase(models.Model):
+    """
+    This model is used to store information about every phase of any paticular build.
+    This includes information such as when the phase was created, started and so on.
```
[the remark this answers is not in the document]
dharmit: I'm not able to understand if this is intended to track only the `docker build` part or the entire build (lint->build->test->scan).

container_pipeline/workers/base.py
```diff
@@ -27,6 +32,49 @@ def handle_job(self, job):
...
+        next_phase, created = BuildPhase.objects.get_or_create(
+            build=self.build, phase=next_phase_name)
+        next_phase.status = 'queued'
+        next_phase.save()
+
+    def set_build_data(self, build_status=None,
+                       build_end_time=None, build_trigger=None):
+
+        if build_status:
+            self.build.status = build_status
```
dharmit: This is not the same as `build_status` we have in the `job`, right? The one in `job` dictionary is a bool.

navidshaikh: correct.

container_pipeline/models/pipeline.py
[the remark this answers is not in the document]
navidshaikh: line no 100-105 should make the confusion clear.

container_pipeline/models/pipeline.py
```diff
@@ -0,0 +1,209 @@
...
+    paticular project on the service. This includes information like its state,
+    when it was created/completed, where are the logs and so on.
+    """
+
+    uuid = models.CharField(max_length=100, unique=True, db_index=True)
+    project = models.ForeignKey(Project)
+
+    STATUS_CHOICES = (
+        ('queued', 'Queued'),
+        ('processing', 'In process'),
```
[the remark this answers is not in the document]
navidshaikh: just an option, ignore the `In process` option

bamachrn:
@mohammedzee1000 please take a look for changing the Database name also, as discussed

container_pipeline/models/pipeline.py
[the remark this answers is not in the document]
dharmit: So this is meant to track all the phases from the start to the end of job? This came up when @mohammedzee1000 explained the flow; can we name it something else? `BuildPhase` sounds to me like something that's got to do with only the build part. 

@bamachrn @mohammedzee1000 can you provide your inputs?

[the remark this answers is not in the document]
mohammedzee1000: There are 2 classes 'Build' and 'BuildPhase'. 'Build' tracks the overall build, while 'BuildPhase' looks at only one particular phase. 

So If a build happens, there will be one entry in 'Build', and for now, upto 5 entries on 'BuildPhase'.

One entry will have phase as dockerlint, while another has phase as build and so on.

@dharmit i hope that answers your question.

mohammedzee1000:
@bamachrn I am already working on it and will hopefully have something up by tommorow.
It required a bit of work though, such as setting up [Database Routers](https://docs.djangoproject.com/en/1.11/topics/db/multi-db/#using-routers)

container_pipeline/models/pipeline.py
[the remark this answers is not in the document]
dharmit: @mohammedzee1000 yep it does. Thank you.

dharmit:
@mohammedzee1000 @bamachrn @navidshaikh 

Summarizing what we discussed on the call. 

Instead of going with two database approach, we will go with one database being shared between Repo Tracking and Data Store. They both are two pieces of the same project - container pipeline service - and there's quite a bit of data that's same for both. 

Immediate consequence of this would probably be that Repo Tracking would start failing/misbehaving at a few points. So, we test with single DB in pre-prod and make sure to fix Repo Tracking wherever needed. Going with two DBs doesn't sound good from architecture point of view.

navidshaikh:
#dotests

mohammedzee1000:
@navidshaikh i think this will require and re base and fixing merge conflicts. I dont want to break this until the call is done

mohammedzee1000:
#dotests

mohammedzee1000:
#dotests-debug

mohammedzee1000:
@bamachrn @dharmit do you guys have any more comments here.

PR #359, opened 2017-10-25
Refactor openshift logs retirval and project deletion

bamachrn:
We have changed the service flow to delete the openshift project after we send the mail to user. This is to make sure we are cleaning up the openshit project even if the build fails, after we collect all the logs for notification.
This PR fixes below things
* send_mail  script path for sending mail with proper from id
* re-factored the provision flow to use same place  in all the machines to store code base
* fixes openshift logs retrieval error due to extra parameters passed
* project deletion after the mail is sent to user


bamachrn:
#dotests-debug

dharmit:
Tested this PR with couple of builds and it looks good. Hence merging.

PR #361, opened 2017-10-27
Stop setting action to <worker>_failure when a worker fails

dharmit:
Linter and delivery worker had some anomalies in that they used a different version of `job` data that was being put to the beanstalkd tube. This PR fixes it by making sure that these workers use global job data created in `pipeline.py` at the beginning of the lint-build-test-scan-deliver process.


bamachrn:
#dotests-debug

PR #362, opened 2017-10-30
Set build_status to false if test or delivery phase fails

dharmit:
We decided to have a status flag for every phase. So, lint -> `lint_status`, build -> `build_status`, so on and so forth. But at the moment, mail server worker doesn't have a classification among these. So if build, test or delivery phases fail, we set `build_status` to fail to inform the user that build has failed and image won't be available on registry.centos.org.

There's already a card on OSIO about refactoring mail worker that @bamachrn is looking into.

### DIFF
--- a/container_pipeline/workers/delivery.py
+++ b/container_pipeline/workers/delivery.py
@@ -33,7 +33,6 @@ class DeliveryWorker(BaseWorker):
         if success:
             self.handle_delivery_success()
         else:
-            self.job["build_status"] = False
             self.handle_delivery_failure()
 
     def deliver_build(self):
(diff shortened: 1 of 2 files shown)


dharmit:
#dotests-debug

bamachrn:
LGTM 👍

container_pipeline/workers/delivery.py
```diff
@@ -33,6 +33,7 @@ def handle_job(self, job):
         if success:
             self.handle_delivery_success()
         else:
+            self.job["build_status"] = False
             self.handle_delivery_failure()
```
navidshaikh: move the flag setting in `self.handle_delivery_failure()` method

container_pipeline/workers/test.py
```diff
@@ -80,6 +80,7 @@ def handle_job(self, job):
         if success:
             self.handle_test_success()
         else:
+            self.job["build_status"] = False
             self.handle_test_failure()
```
navidshaikh: move the flag setting in `self.handle_delivery_failure()` method

PR #363, opened 2017-11-07
Fixes phase name while logging error message in log file

navidshaikh:
This changeset fixes openshift logs retrieval method to log proper phase
  name while there is an issue retrieval of logs. Earlier, if delivery phase
  logs retrieval fails, it will log "build" as phase name. With this fix,
  build, delivery and test phase names are given to log right phase names.

### DIFF
--- a/container_pipeline/lib/openshift.py
+++ b/container_pipeline/lib/openshift.py
@@ -192,7 +192,7 @@ class Openshift(object):
                 empty_retry_count = 0
             time.sleep(retry_delay)
 
-    def get_build_logs(self, project, build_id):
+    def get_build_logs(self, project, build_id, build_type="build"):
         try:
             output = run_cmd(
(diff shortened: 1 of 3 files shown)


navidshaikh:
@mohammedzee1000 @bamachrn : Can you review this?

container_pipeline/lib/openshift.py
```diff
@@ -192,7 +192,7 @@ def wait_for_build_status(self, project, build_id, status,
                 empty_retry_count = 0
             time.sleep(retry_delay)
 
-    def get_build_logs(self, project, build_id):
+    def get_build_logs(self, project, build_id, build_type="build"):
```
navidshaikh: > What about the build worker, i dont see that getting changed.

@mohammedzee1000 : That's default. ;-)

PR #364, opened 2017-11-09
Modify database entry when entry in container-index changes

dharmit:
This PR ensures that when an entry is deleted or modified in container-index, the underlying entry in repo tracking database is deleted. In case of modification to container-index (eg: change in value of `desired_tag`), it deletes the old entry and creates new one.

### DIFF
--- a/jenkinsbuilder/cccp_index_reader.py
+++ b/jenkinsbuilder/cccp_index_reader.py
@@ -159,7 +159,7 @@ def export_new_project_names(projects_names):
         logger.error("Failed to export project names to file %s" %
                      projects_list)
         logger.error("I/O Error {0}:{1}".format(e.errno, e.strerror))
-    except:
+    except Exception as e:
         logger.error("Failed to export project names to file %s" %
                      projects_list)
         logger.error("Unexpected error:%s", sys.exc_info()[0])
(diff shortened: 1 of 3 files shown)


dharmit:
GitHub interface shows I added a lot of things. But actually, most functions were existing and I haven't touched them even a bit. Guess I should prepare myself to get review on the code that I didn't add/modify. :smile:

jenkinsbuilder/cccp_index_reader.py
```diff
@@ -183,52 +185,36 @@ def get_old_project_list():
...
+        new_projects_names.append(
+            str(project[0]["project"]["appid"]) + "/" +
+            str(project[0]["project"]["jobid"]) + ":" +
+            str(project[0]["project"]["desired_tag"])
+        )
+    return new_projects_names
 
 
-def run_command(command):
+def create_or_update_project_on_jenkins(indexdlocation):
```
navidshaikh: just append the new functions to the script, if we insert new functions in between it's confusing and difficult to review.

navidshaikh:
@dharmit : PTAL https://ci.centos.org/job/centos-container-pipeline-service-ci-pr/1358/console , this is console output of CI ran for this PR.

PR #348, opened 2017-10-03
Add summary about scan results

dharmit:
This PR aims to add summary about scan results to the emails being sent to developers. This summary shows up as a one-line comment above the "Detailed logs link" for each scanner:

```
CentOS Community Container Pipeline Service <https://github.com/centos/container-index>                                                                                      
=================================================================                                                                                                            
                                                                                                                                                                             
Build status:           Success                                                                                                                                              
Image:                  registry.centos.org/devcloud/wget:latest                                                                                                             
Build logs:             https://registry.centos.org/pipeline-logs/YTA1NDI3N2NjNT                                                                                             
Cause of build:         Manually triggered by admin                                                                                                                          
                                                                                                                                                                             
---------------------------------------------------------------------------------------------------                                                                          
Atomic scanners results for built image registry.centos.org/devcloud/wget:latest                                                                                             
                                                                                                                                                                             
container-capabilities-scanner:                                                                                                                                              
Container image has few additional capabilities.                                                                                                                             
Detailed logs link: https://registry.centos.org/pipeline-logs/YTA1NDI3N2NjNT/container-capabilities-results.json                                                             
                                                                                                                                                                             
scanner-rpm-verify:                                                                                                                                                          
No issues. Libraries and Binaries are intact.                                                                                                                                
Detailed logs link: https://registry.centos.org/pipeline-logs/YTA1NDI3N2NjNT/RPMVerify_scanner_results.json                                                                  
                                                                                                                                                                             
misc-package-updates:                                                                                                                                                        
No updates for packages installed via pip. No updates for packages installed via npm. No updates for packages installed via gem                                                  
Detailed logs link: https://registry.centos.org/pipeline-logs/YTA1NDI3N2NjNT/misc_package_updates_scanner_results.json                                                       
                                                                                                                                                                             
pipeline-scanner:                                                                                                                                                            
RPM updates available for the image.                                                                                                                                         
Detailed logs link: https://registry.centos.org/pipeline-logs/YTA1NDI3N2NjNT/pipeline_scanner_results.json
```

PR #367, opened 2017-11-14
Scanners to include summary of the results

dharmit:


dharmit:
I've broken #348 into two. This is the first part wherein we're modifying the scanner images only. We're doing this because CI pulls images from registry.centos.org and that would cause failures on CI when we modify scanner images and scan worker in the same PR.

dharmit:
Closed since the commits I rebased off were modifying the files that we don't want to be changing in first part.

PR #368, opened 2017-11-15
Temporary disable tracking for CI check

dharmit:


bamachrn:
@dharmit is it the jenkins job `repo-update-checker` is causing this?

dharmit:
I've removed repo tracking from the CI check this time. Earlier we removed it from the service flow which would not impact the CI flow, IMHO. I'd like to see how things go with it being removed only from CI. If it passes, I think, we can isolate that the way tracking is running in CI is not 100% accurate.

dharmit:
Closing this in favor of #356 which is carrying same commit to disable tracking in CI.

PR #369, opened 2017-11-16
Add scan summary in the output generated by scanners

dharmit:
This PR will make sure that the scanners generate summary about the check and add it to the output. It's not modifying the workers. So, the resulting mail will still not contain any summary. That will be handled in the next PR.

We separated these tasks because in CI, atomic scanner images are pulled from r.c.o and not built during the check. So, if a scanner worker is accessing key `Summary`, it's likely going to fail because the output generated by scanner (based on r.c.o images) won't have the key-value pair.

### DIFF
--- a/atomic_scanners/container-capabilities-scanner/scanner.py
+++ b/atomic_scanners/container-capabilities-scanner/scanner.py
@@ -81,7 +81,7 @@ def template_json_data():
         # "Docker run command": "docker run -d {} tailf /dev/null".format(
         #     IMAGE_NAME),
         "Reference documentation": "http://www.projectatomic.io/blog/2016/01/how-to-run-a-more-secure-non-root-user-container/",
-        "Summary": None
+        "Summary": ""
     }
     return json_out
 
--- a/atomic_scanners/container-capabilities-scanner/scanner.py
+++ b/atomic_scanners/container-capabilities-scanner/scanner.py
@@ -53,7 +53,7 @@ def check_image_for_run_label(image):
     run_label = run_object.get_label("RUN")
 
     if run_label == "":
-        raise RunLabelException("Image doesn't have a RUN label")
+        raise RunLabelException("Dockerfile for the image doesn't have RUN label")
     return run_label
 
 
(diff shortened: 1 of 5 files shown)


dharmit:
Here's the `grep` output for an image that contained outdated packages installed via yum, pip, gem and npm. Other scanners also have appropriate summaries:

```
# grep -r Summary /srv/pipeline-logs/NzYxMzYzN2M2Mm
/srv/pipeline-logs/NzYxMzYzN2M2Mm/RPMVerify_scanner_results.json:        "Summary": "Libraries and binaries in the image are intact.", 
/srv/pipeline-logs/NzYxMzYzN2M2Mm/container-capabilities-results.json:            "Summary": "Dockerfile for the image doesn't have RUN label.", 
/srv/pipeline-logs/NzYxMzYzN2M2Mm/misc_package_updates_scanner_results.json:            "Summary": "Possible updates for packages installed via pip. ", 
/srv/pipeline-logs/NzYxMzYzN2M2Mm/misc_package_updates_scanner_results.json:            "Summary": "Possible updates for packages installed via npm. ", 
/srv/pipeline-logs/NzYxMzYzN2M2Mm/misc_package_updates_scanner_results.json:            "Summary": "Possible updates for packages installed via gem. ", 
/srv/pipeline-logs/NzYxMzYzN2M2Mm/pipeline_scanner_results.json:        "Summary": "RPM updates available for the image.",
```

atomic_scanners/container-capabilities-scanner/scanner.py
```diff
@@ -53,7 +53,7 @@ def check_image_for_run_label(image):
     run_label = run_object.get_label("RUN")
 
     if run_label == "":
-        raise RunLabelException("Image doesn't have a RUN label")
+        raise RunLabelException("Dockerfile for the image doesn't have RUN label.")
```
navidshaikh: Good.

atomic_scanners/container-capabilities-scanner/scanner.py
```diff
@@ -80,10 +80,12 @@ def template_json_data():
         "Scan Results": {"Container capabilities": None},
         # "Docker run command": "docker run -d {} tailf /dev/null".format(
         #     IMAGE_NAME),
-        "Reference documentation": "http://www.projectatomic.io/blog/2016/01/how-to-run-a-more-secure-non-root-user-container/"
+        "Reference documentation": "http://www.projectatomic.io/blog/2016/01/how-to-run-a-more-secure-non-root-user-container/",
+        "Summary": None
```
navidshaikh: Summary key in dict will hold text value, as a point of having uniform values a key can get, a default value equivalent to `None` could be `""`.

navidshaikh:
@dharmit : is this ready to merge ?

dharmit:
I believe it is ready. We need one more review.
 
cc: @bamachrn @mzee1000

bamachrn:
LGTM 👍 
this is for scanner images and we are putting worker updates in a different PR

PR #371, opened 2017-11-23
Changed pre-build job definition and renderer

bamachrn:
* pre-build job generationupdated as jenkins is not able to read file from out side of workspace
* updated jenkins job config for enabling git push to rendered repository

### DIFF
--- a/pre_build/cccp_index_pre_build_check.py
+++ b/pre_build/cccp_index_pre_build_check.py
@@ -7,10 +7,10 @@ Once entries with prebuild_script parameter found, this class generates one job 
 Then it runs the jenkins-job builder command for creating the job in ci.centos.org jenkins.
 """
 
-import os
+import random
+import string
 import subprocess
 import sys
-import tempfile
 from glob import glob
 
 import yaml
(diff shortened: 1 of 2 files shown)


pre_build/cccp_index_pre_build_check.py
```diff
@@ -116,48 +116,44 @@ def main(indexdlocation):
...
-                t = tempfile.mkdtemp()
-                generated_filename = os.path.join(
-                    t,
-                    'cccp_GENERATED.yaml'
-                )
+                pre_text = ''.join(random.sample(
+                    string.lowercase + string.digits, 10))
+                generated_filename = '_'.join(pre_text,
+                                              'cccp_GENERATED.yaml'
+                                              )
```
navidshaikh: so `randomString_cccp_GENERATED.yml` is going to be inside workspace directory and not in temporary directory?

pre_build/pre-build-job.yml
```diff
@@ -57,18 +54,22 @@
...
-              "git add -A && "
-              "git commit -am 'Added changes from pre-build' && "
-              "git push prebuild-render {prebuild_branch} && "
+              $ssh_cmd -t "cd payload && rm -rf .gitignore .git .gitkeep||true && \
+              git init && git config --global user.name 'Container Pipeline' && \
+              git config --global user.email container@centos.org && \
+              git remote add prebuild-render {prebuild_render} && \
+              git checkout -b {prebuild_branch} && \
+              git add -A && \
+              git commit -am $GIT_COMMIT && \
```
navidshaikh: what will be the value of `$GIT_COMMIT` or where it is declared/defined?

pre_build/cccp_index_pre_build_check.py
bamachrn: yup, and every run cleans the workspace.

pre_build/pre-build-job.yml
bamachrn: this is jenkins variable commit id from the source repo. we can make use of it for build reason in next iteration

PR #375, opened 2017-11-24
[WIP] Clean up ansible tasks for database

navidshaikh:
This changeset has clean up ansible tasks for removing artifacts of database, for ansible clean up utility of service.

### DIFF
--- a/provisions/roles/cleanup/jenkins/master/tasks/main.yml
+++ b/provisions/roles/cleanup/jenkins/master/tasks/main.yml
@@ -11,3 +11,10 @@
   sudo: yes
   ignore_errors: yes
   shell: docker rmi $(docker images -f "dangling=true" -q)
+
+- name: Stop and remove postgresql container
+  sudo: yes
+  ignore_errors: yes
+  docker_container:
+    name: postgres
+    state: absent


navidshaikh:
This is ready to review and merge.

dharmit:
Naive question: how does stopping and removing postgresql container help?

In the OSIO card related to this PR I see:
> The container should be stopped and removed as part of clean up utility during deployments.

But, why exactly do we need to remove it?

navidshaikh:
> Naive question: how does stopping and removing postgresql container help?
Because
1. We want a fresh deployment. 
2. Given persistent data dir as volume, restarting the container should ensure data recovered.
3. Given any changes made to container start parameters, ensure applied changes.
4. In line with, removing and starting worker containers.

dharmit:
> Given persistent data dir as volume, restarting the container should ensure data recovered.

What does data recovery have to do with stopping -> removing -> starting new postgresql container?

> Given any changes made to container start parameters, ensure applied changes.

👍

navidshaikh:
> What does data recovery have to do with stopping -> removing -> starting new postgresql container?

Just to be verbose and addressing concern if stopping and removing the container may loose data.

dharmit:
> Just to be verbose and addressing concern if stopping and removing the container may loose data.

So, are we doing some check for this? Are we doing anything that ensures that when cleanup job deletes the container and doing a fresh provision starts a new container, the data remains intact? 

Besides, what if we clean this container at the time of cleanup and provision fails for some reason? Then, what happens to the UI piece that uses Django and interacts with DB to deliver the data about builds to users?

> > Given any changes made to container start parameters, ensure applied changes.

> 👍

I believe this kind of change is going to happen infrequently. And when it does happen, we can probably ensure that container is removed manually before a provision is made.

> In line with, removing and starting worker containers.

I don't entirely get the rationale here.

navidshaikh:
|what happens to the UI piece that uses Django and interacts with DB to deliver the data about builds to users?

Do we want to block a current work based on something which is not in place yet?

I think its responsibility for UI's work to ensure the data is always available.

| Are we doing anything that ensures that when cleanup job deletes the container

What kind of check do you propose? Do we have check after writing ansible tasks for creating the container that indeed the container is created?

>  and doing a fresh provision starts a new container, the data remains intact?
That needs to be tested, the data-store PR is being tested in pre-prod environment. 

I should add [WIP] tag to the PR instead.

dharmit:
> Do we want to block a current work based on something which is not in place yet?

Well, no. But, you see the connection and rework that's coming up if we merge this. Besides, we're getting close to having UI in place.

> I think its responsibility for UI's work to ensure the data is always available.

👎 
how is it front-end's ***responsibility*** to ensure that back-end is always available? Enlighten me, please. Am I missing something here @bamachrn?

> What kind of check do you propose?

You're proposing "data recovery" as a reason for restarting the container. Not me. This is not even "data recovery". Are we trying to restore database using a backup? That would qualify as "data recovery".

> Do we have check after writing ansible tasks for creating the container that indeed the container is created?

Yes, Ansible will return an error if the container if it fails to create container for any reason and the playbook execution stops at that place unless you've asked Ansible to ignore errors for that task.

dharmit:
@bamachrn since when did we decide to add "WIP" as a label to a PR and not to the heading of the PR?

bamachrn:
I have few points here.
1. sole purpose of the storing data in database is to make that data available to the user/developer. now the data can be available in three ways.
   -  through email if we want put some tracking data or other details from database.
   -  through UI, wehere we want to show the status,logs, dockerfiles as early as poissible (aim is to show this real time)
   - through some REST API in case developer wants to debug on something.

2. We have broadly two pieces for our team
   - registry.centos.org, which is meant to be always up and available for the users to pull images from
   - container pipeline service, which builds the images from the container-index and push it to registry. now this service can be down for few minutes (in case of emargency).

in this PR. we are cleaning up the postgres container, which would be having reference from registry UI. And anything which might break registry (or makes registry ui unavailable), I would say we should avoid it.

Now this has few things to take care also.
1. we should be updating the database as required, but while updating we should make sure we are updating the registry UI also respectively. making sure registry is not down/broken in any case.
2. whenever we are putting something as back bone (in our case it the database we are introducing), we should always foresee the scenarios which are going to bite us.
    - if we say, we would stop the availability of data and let the UI findout how it would fetch data for showing to user. it does not make any sense at all.  Rather, database should always serve the data as requested by UI, even if the service is down

So as my proposal for refreshing the pgsql container, I would say lets do it while provisioning it self. if we change any configuration for a running container, ansible already takes care of restarting it. We should never clean it up in the service cleanup process and wait for the deployment to complete. as this means we are gambling on availability of data to user.

@navidshaikh @dharmit HDYT about this?

navidshaikh:
Closing the PR since UI needs data served from container.

PR #376, opened 2017-11-24
Fixes configurations for postgresql database container

navidshaikh:
This changeset fixes configurations for postgresql container for data volume.

### DIFF
--- a/provisions/group_vars/all
+++ b/provisions/group_vars/all
@@ -20,7 +20,6 @@ db_pass: cccp
 db_host: '0.0.0.0'
 db_port: 5432
 allowed_hosts: "['127.0.0.1']"
-db_backup_host_path: "/srv/pipeline-logs/db/cccp/"
 
 sentry_enabled: false
 sentry_log_level: 'WARNING'
--- a/provisions/roles/db/tasks/main.yml
+++ b/provisions/roles/db/tasks/main.yml
@@ -3,6 +3,7 @@
   file: path= "{{ db_backup_host_path }}" state=directory
   sudo: yes
   tags: db
+  when: not test
 
 - name: Set selinux context for datadir for mounting in containers
   sefcontext:
--- a/provisions/roles/db/tasks/main.yml
+++ b/provisions/roles/db/tasks/main.yml
@@ -43,6 +43,23 @@
       restart_policy: on-failure
   sudo: yes
   tags: db
+  when: not test
+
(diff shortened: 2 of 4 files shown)


provisions/group_vars/all
```diff
@@ -20,7 +20,6 @@ db_pass: cccp
 db_host: '0.0.0.0'
 db_port: 5432
 allowed_hosts: "['127.0.0.1']"
-db_backup_host_path: "/srv/pipeline-logs/db/cccp/"
```
mohammedzee1000: I believe this can fails if we dont have this value. The CI inventory has specified the value, does not mean its there in any other inventory file, unless we want to make it compulsory in inventory file.

navidshaikh:
#dotests-debug

navidshaikh:
#dotests-debug

provisions/roles/db/tasks/main.yml
```diff
@@ -1,12 +1,13 @@
 ---
 - name: Ensure cccp data dir exists
-  file: path=/var/lib/cccp/db state=directory
+  file: path= "{{ db_backup_host_path }}" state=directory
   sudo: yes
   tags: db
+  when: not test
```
dharmit: Does this mean that when we're doing a CI deployment, the backup directory won't be created? But it'll still be created in DevCloud, right? How about making it something like `when: production` so that it's created only for prod? We don't need it for either DevCloud or CI.

provisions/roles/db/tasks/main.yml
```diff
@@ -1,12 +1,13 @@
...
 
 - name: Set selinux context for datadir for mounting in containers
   sefcontext:
-      target: '/var/lib/cccp/db(/.*)?'
+      target: '{{ db_backup_host_path }}'
```
dharmit: IIUC, in CI environment, we don't create the backup directory. Won't this task fail because we're trying to SELinux context to a dir that doesn't exist?

provisions/roles/db/tasks/main.yml
navidshaikh: @dharmit : But for testing purpose we need to make sure the database also works, the database is going to have mutiple functionality on top of it, like APIs UI etc, the host directory availability for the service should be less/no-overhead to developer, it should be taken care by provisioning itself.

bamachrn:
LGTM for now. we need to look for the NFS setup in next iteration

PR #378, opened 2017-11-24
Updating index validation to check prebuild script.

mohammedzee1000:
This also removes the check for target file, if
a prebuild script is specified as in case of
prebuild, the target file is unlikely to already
exist, and will instead be generated as artifact
during prebuild.

### DIFF
--- a/testindex/validators.py
+++ b/testindex/validators.py
@@ -122,6 +122,7 @@ class IndexFormatValidator(IndexValidator):
                     id_list.append(entry["id"])
 
             # Check for pre-build script
+            # TODO: This should be updated soon to fail if prebuild-script is not in entry and warn if it is None
             if "prebuild-script" not in entry or ("prebuild-script" in entry and entry["prebuild-script"] is None):
                 self._summary_collector.add_warning("No pre-build script specified.")
 
@@ -320,11 +321,12 @@ class IndexProjectsValidator(IndexValidator):
 
             # * Check for pre-build script
             # TODO : Make a better implementation of pre-build script checking
+            # TODO : Ideally, if prebuild is not in entry, it wont reach here and this should happen if it is not None
             prebuild_exists = False
             if "prebuild-script" in entry:
                 prebuild_exists = True
                 prebuild_script = entry.get("prebuild-script")
-                if prebuild_script and not path.exists(git_path + "/" + str(prebuild_script)):
+                if prebuild_script and not path.exists(path.join(git_path, str(prebuild_script))):
                     self._mark_entry_invalid(entry)
                     self._summary_collector.add_error("Invalid pre-build script path specified")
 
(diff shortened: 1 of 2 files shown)


testindex/validators.py
```diff
@@ -121,6 +121,10 @@ def run(self):
                 else:
                     id_list.append(entry["id"])
 
+            # Check for pre-build script
+            if "prebuild-script" not in entry or ("prebuild-script" in entry and entry["prebuild-script"] is None):
+                self._summary_collector.add_warning("No pre-build script specified.")
```
navidshaikh: 1. This will add warning for all the projects in index, which doesn't specify `prebuild-script`.
2. If checking for format of entry having `pre-build-script`, `add_error`should be used for `None` value of `prebuild-script` key.
3. Make it simple.

testindex/validators.py
```diff
@@ -314,8 +318,18 @@ def run(self):
 
             container_names[container_name].append(entry["id"])
 
+            # * Check for pre-build script
+            # TODO : Make a better implementation of pre-build script checking
+            prebuild_exists = False
+            if "prebuild-script" in entry:
+                prebuild_exists = True
+                prebuild_script = entry.get("prebuild-script")
+                if prebuild_script and not path.exists(git_path + "/" + str(prebuild_script)):
```
navidshaikh: us `os.path.join` while concatenating paths

testindex/validators.py
mohammedzee1000: @navidshaikh shaikh true. It will add a warning. But warning does not fail the validation.

 It should infact be a failure, but i am making it a warning for now, so it does not fail validation. We will first need update all index entries to include the new field, before i turn it into a failure.

This goes along the lines of @kbsingh ,wanting all entries in the index to be explicitly specified, even if it is None.

This brings us to the second point of what if a user does not have a prebuild-script. Then he should say None, and that goes out as a warning that it is not being used. 

The idea being index validation warning is its ok to have them, so long as there is valid justification. But people should be able to make that informed decision, before merging the PR.

mohammedzee1000:
#dotests

testindex/validators.py
```diff
@@ -314,8 +323,21 @@ def run(self):
...
             container_names[container_name].append(entry["id"])
 
+            # * Check for pre-build script
+            # TODO : Make a better implementation of pre-build script checking
+            # TODO : Ideally, if prebuild is not in entry, it wont reach here and this should happen if it is not None
+            prebuild_exists = False
+            if "prebuild-script" in entry and "prebuild-path" in entry:
+                prebuild_exists = True
+                prebuild_script = entry.get("prebuild-script")
+                prebuild_path = entry.get("prebuild-path")
```
bamachrn: @mohammedzee1000 can we change `prebuild_path` to `prebuild_context`? this will give the user a better reason for adding the variable

mohammedzee1000: @bamachrn sure, i can do that.

PR #381, opened 2017-12-01
Adds inventory file patterns in .gitignore

navidshaikh:
With different deployment environments, we have different inventory files,
  adding the patterns in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,9 @@
 *.pyc
 .vagrant
 *.sqlite3
+*.retry
+hosts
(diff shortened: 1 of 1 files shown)


dharmit:
I assume that putting these files in `.gitignore` be required when we do deployments from dev system. I've always deployed the service to DevCloud and never happened to encounter these files in `git status`.

But yeah if others have faced this issue, I don't see why we can't review and merge this.

@bamachrn @mohammedzee1000 can you guys review this?

PR #382, opened 2017-12-01
Documents utilities in library script for CI

navidshaikh:
This changeset documents library utilities and fixes pep8 issues.


ci/lib.py
```diff
@@ -87,29 +93,45 @@ def read(out, err):
 
 
 class ProvisionHandler(object):
+    """
+
```
dharmit: Why the empty line?

navidshaikh:
> Can we have a short description about the parameters that a function takes and what it returns 

@dharmit : done

PR #383, opened 2017-12-01
Adds create_project function to linter worker

dharmit:
Linter worker creates a new project on OpenShift when the lint proces
is completed successfully. Instead of having it in a separate file, it
helps to have the function in the same file.


container_pipeline/workers/linter.py
```diff
@@ -3,19 +3,90 @@
...
+    """
+    Creates a new project in OpenShift. This function expects a queue on which
+    the job is to be put, the job itself and logger to be used.
+
+    :param queue: beanstalkd queue on which job is to be put
+    :param job: job to be put on the beanstalkd queue
+    :param logger: logger to be used to log things
+    :type queue: str
+    :type job: dict
+    :logger: logging.logger
```
navidshaikh: +1 for this

dharmit:
Created a fresh DevCloud environment to run CI check and figure what's causing the failure.

dharmit:
#dotests

dharmit:
#dotests

provisions/roles/jenkins/slave/templates/template.json.j2
```diff
@@ -86,7 +86,8 @@
     },
     {
       "name": "BUILD_CONTEXT",
-      "description": "The build context for the container build"
+      "description": "The build context for the container build",
+      "value": "."
```
navidshaikh: should this be `./` to match with the defaults we have used across for this field ?

dharmit: I used `"."` based on [this](https://github.com/CentOS/container-pipeline-service/blob/master/jenkinsbuilder/cccp_index_reader.py#L113). If this commit helps make CI green, we can take @mohammedzee1000's opinion about what to set it as.

I don't think this commit should really help clear the CI, however. How is it that `master` branch is working just fine? 😞

navidshaikh: and I was talking about this https://github.com/CentOS/container-index/pull/234/files

Also the CI is green now. :-D

mohammedzee1000: Lets go ahead with it then :)

bamachrn:
LGTM,merging

PR #386, opened 2017-12-06
Separates execution and exception blocks for operations in CI

navidshaikh:
This changeset aims to separate the tasks needed for CI run, viz
 [get_nodes, deploy+build, test, teardown].
 With this in place, we can catch and report issues happened in particular
 task and do not print unnecessary data on CI console.

### DIFF
--- a/ci/cccp_ci.py
+++ b/ci/cccp_ci.py
@@ -114,9 +114,9 @@ if __name__ == '__main__':
     except Exception as e:
         _print('Build failed in either deployment or running builds: %s' % e)
         # TODO: cat deployment logs
-        _if_debug()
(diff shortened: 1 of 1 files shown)


navidshaikh:
Check console [output](https://ci.centos.org/job/centos-container-pipeline-service-ci-pr/1469/console) to verify if changes work as expected. Note that, we have intentionally failed a test to ensure we do not cat the log file.

bamachrn:
this looks good, here we are exiting if the previous step fails right?

navidshaikh:
> here we are exiting if the previous step fails right?

that's correct, and printing only relevant information for failed step.

PR #387, opened 2017-12-06
Uses local dir as data dir for postgresql container

navidshaikh:
Use local directory as data dir with postgresql container and set right ACL for same.

This change-set also parameterize the uid for postgresql container. Default is `26` as mentioned in postgresql [Dockerfile](https://github.com/sclorg/postgresql-container/blob/master/9.5/Dockerfile#L68). 

The uid can be overridden while creating container, and thus one can change the value of same in inventory file.

Also updates the default path given in inventory file to `/srv/db/cccp/` since, earlier path default path was mistakenly given in logs NFS path.

### DIFF
--- a/ci/lib.py
+++ b/ci/lib.py
@@ -232,7 +232,6 @@ scanner_worker
 {scanner_host}
 
 [all:vars]
-db_host= {jenkins_master_host}
 public_registry= {jenkins_slave_host}
 intranet_registry = {jenkins_slave_host}:5000
 copy_ssl_certs=True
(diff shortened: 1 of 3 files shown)


dharmit:
@navidshaikh I understand that we're going to rsync the data to NFS storage. Do we have an automated (`cron`?) `rsync` job as a part of Ansible deployment? I think, no. Are we going to open a new PR for it or do we have a PR open already?

navidshaikh:
Closing in favor of https://github.com/CentOS/container-pipeline-service/pull/400

PR #388, opened 2017-12-06
Stops printing command ran on CI console

navidshaikh:
This to minimize the unnecessary text printed on CI console.

### DIFF
--- a/ci/lib.py
+++ b/ci/lib.py
@@ -42,6 +42,8 @@ def run_cmd(cmd, user='root', host=None, private_key='', stream=False):
     Note:
         If stream=True, the function writes to stdout.
     """
+
+    """ # NOTE: Stop printing run command on CI console, uncomment if needed
     _print('=' * 30 + 'RUN COMMAND' + "=" * 30)
     _print({
         'cmd': cmd,
@@ -50,6 +52,7 @@ def run_cmd(cmd, user='root', host=None, private_key='', stream=False):
         'private_key': private_key,
         'stream': stream
     })
+    """
     if host:
         private_key_args = ''
         if private_key:


dharmit:
I've been doing CI debugging lately for a few PRs that fail CI check. And it's really helpful to have info about the command that failed CI check. I believe that with this changeset, we stop logging the command that fails. 

I vote to not merge this PR. While I agree that we need to reduce the amount of logs being printed on console, I vote against going ahead with this particular one.

navidshaikh:
> And it's really helpful to have info about the command that failed CI check.

@dharmit 
if you check, https://github.com/CentOS/container-pipeline-service/blob/master/ci/lib.py#L111 , we are printing the command if it there are any errors in executing it.

navidshaikh:
@dharmit 

Some more to help understand, we do print the command failed in CI

```
Exception: Run Command Error for: ssh -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no  root@n6.gusty 'cat /tmp/deploy.logs'
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added 'n6.gusty,172.19.2.134' (ECDSA) to the list of known hosts.
cat: /tmp/deploy.logs: No such file or directory
```

from https://ci.centos.org/job/centos-container-pipeline-service-ci-pr/1473/console

dharmit:
@navidshaikh thanks. As long as we're printing the command that failed along with why it failed, I'm fine. 👍 

@bamachrn can you take a look once?

bamachrn:
@navidshaikh IIUC this PR is for making sure we are not printing the commands before running them. and if some command fails it is printed in the exception right?

bamachrn:
wow! this is merged.

navidshaikh:
> if some command fails it is printed in the exception right?

Right. Also mentioned in earlier comments https://github.com/CentOS/container-pipeline-service/pull/388#issuecomment-350221937

PR #389, opened 2017-12-06
Export deploy logs in a file, cat only if deploy fails

navidshaikh:
To have better CI console logs, only needed information is exported.

### DIFF
--- a/ci/cccp_ci.py
+++ b/ci/cccp_ci.py
@@ -100,6 +100,7 @@ if __name__ == '__main__':
     try:
         # get nodes from CICO infra
         nodes = get_nodes(count=5)
+        _print(str(nodes))
     except Exception as e:
         _print('Build failed while receiving nodes from CICO: %s' % e)
         _if_debug()
@@ -114,7 +115,7 @@ if __name__ == '__main__':
     except Exception as e:
         _print('Build failed in either deployment or running builds: %s' % e)
         # first cat the deployment logs, nodes[4] = controller node
-        _print(run_cmd('cat %s' % DEPLOY_LOGS_PATH, host=nodes[4]))
+        _print(run_cmd('cat %s' % DEPLOY_LOGS_PATH, host=nodes[-1]))
         # then cat the cccp.log, nodes[1] = jenkins slave
         _print(run_cmd('cat /srv/pipeline-logs/cccp.log', host=nodes[1]))
         _if_debug()
--- a/ci/cccp_ci.py
+++ b/ci/cccp_ci.py
@@ -12,7 +12,8 @@ import json
 import os
 import sys
 
-from ci.lib import _print, setup, test, teardown, run_cmd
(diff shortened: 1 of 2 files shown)


dharmit:
AFAIK, the problem we have is this: when everything passes the CI, we don't pay much heed to whether console logs are available on Jenkins or not. When something fails, we care about it. But, in case of failure, it's difficult to understand what failed. This is because of the sheer amount/size (in terms of lines) of logs that gets printed.

For eg: one of the PRs that was recently failing CI check had problems with uploading OpenShift template. Exact error for this was not seen anywhere in about four or five CI runs that it went through. When we debugged it, we found that error was getting printed in dockerfile-lint worker's logs. We searched console log but couldn't find the same error.

Is this PR solving that kind of a problem? I don't mean exactly that problem.

Besides that, this PR carries same code as that by #388. 

Also, what's the connection between installing `nfs-utils2` and this PR?

navidshaikh:
> Also, what's the connection between installing nfs-utils2 and this PR?

if you check last commit to PR, you will see that the commit intentionally tries to break CI to have desired structure of logs printed, the commit tries to install a package that does not exist.

navidshaikh:
> Besides that, this PR carries same code as that by #388.

correct, to not increase the PR sizes and PR is rebased onto others wherever needed.

navidshaikh:
> Is this PR solving that kind of a problem? I don't mean exactly that problem.

We can reach to solutions at one go with enormous code in a single PR and prone to bugs or we can decide in going with iterations. We'll go about having meaningful logs in multiple iterations.

ci/cccp_ci.py
```diff
@@ -43,36 +66,68 @@ def get_nodes(ver="7", arch="x86_64", count=4):
...
+    sys.exit(1)
+
+
 if __name__ == '__main__':
     try:
+        # get nodes from CICO infra
         nodes = get_nodes(count=5)
+        _print(str(nodes))
+    except Exception as e:
+        _print('Build failed while receiving nodes from CICO: %s' % e)
```
bamachrn: Can we change the message to something like `could not get node from CICO for setup`. Also I would say lets add one more iteration for getting nodes. if it fails second time, we exit. Because this has got nothing to do with the service code and if we dont get node, CI is going to fail unnecessarily.

navidshaikh: Noted. Will have a separate PR handling this.

navidshaikh:
@dharmit @mohammedzee1000 @bamachrn : PTAL the CI console

For passing CI : https://ci.centos.org/job/centos-container-pipeline-service-ci-pr/1482/console 
For failing CI due to ansible deploy errors: https://ci.centos.org/job/centos-container-pipeline-service-ci-pr/1487/console (or console of last CI run on this PR)

navidshaikh:
@dharmit @bamachrn @mohammedzee1000 : This is ready to review and merge.

PR #391, opened 2017-12-07
Force pull container-capabilities-scanner during deployment

dharmit:
At the time of deployment, we force pull all scanner container images to ensure that they're fully updated. The container-capabilities-scanner image was missing this force pull config and caused us `KeyError: 'Summary'` issue.

This PR fixes the same.


bamachrn:
LGTM, merging this as we need this hot fix for prod

PR #394, opened 2017-12-11
Wait in queue untill the previous build is completed

bamachrn:
Currently we are waiting for 500 seconds for previous job to get finished. Now if there are a few jobs are in queue then it is taking more than this time. Because of which job is finishing with a failed notification sent to user.
In this PR, after waiting for 500 seconds if the previous build is still not completed we are queuing the job back for next try. This will stop the false alarms of the failed build for the cases of openshift project not getting freed from the previous run. 

For example. let say there is `job-1` and `job-2` request for the same project `project-1`
`job-1` is in progress so there is one openshift project for this. This openshift project gets deleted only if the flow for `job-1` completes. 
Now in the meantime `job-2` comes to linter for geting in into the tubes., but this PR makes sure `job-2` is pushed back to the linter tube untill `job-1` finishes.


bamachrn:
closing this PR will reopen after changes

dharmit:
The description for the PR is somewhat tricky to understand. Can you please provide an example of flow of things based on how they'll look after this PR is merged? For eg: if `job-1` is in queue for 500 seconds, we won't kill it anymore because previous jobs might take long to finish.

Makes sense?

container_pipeline/workers/linter.py
```diff
@@ -81,15 +81,37 @@ def lint(self):
             out, err = run_cmd_out_err(command)
             if err == "":
                 response = self.handle_lint_success(out)
+                self.logger.info(
```
dharmit: This logger statement needs to come before we try to do `handle_lint_success` because that method tries to create an OpenShift project. Doesn't make sense to log trying to do something after trying it.

container_pipeline/workers/linter.py
```diff
@@ -81,15 +81,37 @@ def lint(self):
...
+                    "Docker file linting successful going for job creation")
```
dharmit: We're not going for job creation here. We're doing OpenShift project creation.

container_pipeline/workers/linter.py
```diff
@@ -81,15 +81,37 @@ def lint(self):
...
+                    self.logger.info("Deleting 'dockerfile' data from job")
+                    self.job["dockerfile"] = None
+                    self.job["retry"] = None
             else:
                 response = self.handle_lint_failure(err)
                 self.job["dockerfile"] = None
                 self.job["action"] = "notify_user"
-                self.queue.put(json.dumps(self.job), 'master_tube')
+
+            if self.job.get("retry") > self.MAX_RETRY:
```
dharmit: Comparing `None` with integer when we're able to successfully create an OpenShift project in first attempt is going to break things here and we're going to see exception in the logs. Also breaks the flow for particular job, IMHO.

dharmit: Okay, I take it back, it does work:

```python
In [1]: None > 1
Out[1]: False

In [2]: None > 10
Out[2]: False

In [3]: None < 10
Out[3]: True
```

container_pipeline/workers/linter.py
```diff
@@ -81,15 +81,37 @@ def lint(self):
...
+                self.job["dockerfile"] = None
+                self.job["action"] = "notify_user"
+                self.job["notify_email"] = "container-status-report@centos.org"
```
dharmit: Please don't change the notification email here because we get numerous SUCCESS/FAIL emails on this list in a day already. If we change the workflow and send OpenShift project creation failure mails to the list instead of sending to end user, we might end up missing such emails and until someone takes a close look or a user points out that hist/her image hasn't been built since some time, no one might be aware of larger underlying issue(s).

container_pipeline/workers/linter.py
bamachrn: Done

container_pipeline/workers/linter.py
bamachrn: done,

container_pipeline/workers/linter.py
bamachrn: done as discussed

bamachrn:
running a build for email sample. with updated changes

container_pipeline/workers/linter.py
```diff
@@ -80,30 +80,29 @@ def lint(self):
...
             if err == "":
-                response = self.handle_lint_success(out)
                 self.logger.info(
-                    "Docker file linting successful going for job creation")
+                    "Dockerfile linting successful going for openshift job creation")
+                response = self.handle_lint_success(out)
                 if not response["job_created"]:
                     self.logger.warning(
-                        "Openshift project is not created putting it back to tube")
+                        "Openshift project is not created putting it back to linter tube")
```
dharmit: `"not created putting it"` putting what? Please do `s/it/job` or something similar.

bamachrn:
This is working as expected now @dharmit you want to approve this one?

PR #395, opened 2017-12-11
Fix typo for case when cause of build is change in upstream project

dharmit:
PR title says it all. PTAL @bamachrn


bamachrn:
LGTM, merging it

PR #400, opened 2017-12-13
postgresql db container volume fix, cron job for hourly backup

navidshaikh:
- Postgresql cccp db backup by linking another postgresql container
- NFS directory for backups to be mentioned in ansible inventory
- Hourly cron job for backing up the cccp db
- The cron job back up container which runs the pg_dump uses NFS directory for exporting the dump files
- Parameterized more variables 
- sets variable names and default values in sample inventory file
- adds ansible task for ensuring the local volume dir exists
- updated ansible task to use volume
- added new variable postgresql_uid in group vars
- added ansible task to change ACL for local vol dir to postgresq_uid


provisions/group_vars/all
```diff
@@ -19,6 +19,7 @@ db_name: cccp
 db_pass: cccp
 db_host: '0.0.0.0'
 db_port: 5432
+postgresql_image: registry.centos.org/sclo/postgresql-95-centos7
```
mohammedzee1000: +1 for parameterising this

dharmit: Isn't this container image moving to `centos` namespace?

navidshaikh:
@dharmit  @mohammedzee1000 @bamachrn : please review the PR.

provisions/roles/db/tasks/main.yml
```diff
@@ -1,18 +1,32 @@
...
+  register: nfs_dir_mounted
+  when: not test
+  tags: db
+
+- name: Fail if NFS share is not mounted on node for db backup
+  sudo: yes
+  fail:
+    msg: "NFS share is not mounted on node for db backup."
+  when:
+    - not test
```
bamachrn: I did not get what does this job check. how is it coming to conclusion of nfs is not mounted

provisions/roles/db/tasks/main.yml
```diff
@@ -33,11 +47,11 @@
 - name: Run postgres container
   docker_container:
       name: postgres
-      image: registry.centos.org/sclo/postgresql-95-centos7
+      image: '{{ postgresql_image }}'
       volumes:
-          - "{{ db_backup_host_path }}:/var/lib/pgsql/data"
+        - "{{ db_local_volume }}:/var/lib/pgsql/data:Z"
```
bamachrn: 👍

provisions/roles/db/templates/cccp_db_pg_dump_cron.sh.j2
```diff
@@ -0,0 +1,27 @@
...
+function pg_backup_database
+{
+  docker run --rm --link postgres -v $BACKUPDIR:/tmp $PGIMAGE bash -c "$PGDUMPCOMMAND"
+  echo "finished."
+}
+
+echo "------------------------------------------------------"
+echo `date +'%F %T'` Starting postgresql $DBNAME db backup..
+echo "------------------------------------------------------"
+pg_backup_database
```
bamachrn: could you please point me some reference of `pg_backup_database`

provisions/roles/db/tasks/main.yml
navidshaikh: This is to ensure the NFS directory is present on the node for cron job. And we don't need cron job in CI, so this will not run in CI, i.e. when `test=False` in inventory.

provisions/roles/db/templates/cccp_db_pg_dump_cron.sh.j2
navidshaikh: @bamachrn What do you mean by reference of pg_backup_database? It's a bash function defined in same file at line no:16

provisions/roles/db/tasks/main.yml
mohammedzee1000: Nice

mohammedzee1000:
Cool we are good to merge then, merging

PR #401, opened 2017-12-18
[Do not merge]Delete openshift project forcefully

bamachrn:
This PR is for making sure openshift does not fail to delete the openshift project on request. This first checks whether the project is existing or not. If not found it logs as project is not existing. If found it deletes the project.


dharmit:
We're surely interested in making sure that the project deletion happens properly. However, we also saw that `--force` didn't help when the user (here, `test-admin`) was not privileged to delete the project. 

Now this is weird because `test-admin` is *supposed* to be the user that created the project. If the project creator is not able to delete it, this is really weird.

I'd request not to merge this until we're entirely sure. We might not need `--force` at all, if something else is being messed up.

PR #402, opened 2017-12-18
Fixes deleting stale projects at jenkins

navidshaikh:
Ensures the `all_projects_names.txt` file persist across redeployments and is not cleared up by cleanup ansible utilities.

### DIFF
--- a/jenkinsbuilder/cccp_index_reader.py
+++ b/jenkinsbuilder/cccp_index_reader.py
@@ -10,6 +10,7 @@ from glob import glob
 
 import yaml
 
+from container_pipeline.lib.settings import LOGS_BASE_DIR
 from container_pipeline.lib.log import load_logger
 from container_pipeline.model_tmp.containers import (ContainerLinksModel,
                                                      form_Dockerfile_link)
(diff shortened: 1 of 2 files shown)


jenkinsbuilder/cccp_index_reader.py
```diff
@@ -327,9 +328,12 @@ def main(indexdlocation):
...
 
-    # delete stale entries at jenkins
-    logger.debug("Deleting stale projects.")
-    delete_stale_projects_on_jenkins(stale_projects)
+    if not stale_projects:
+        # delete stale entries at jenkins
+        logger.debug("Deleting stale projects.")
+        delete_stale_projects_on_jenkins(stale_projects)
+    else:
+        logger.info("No stale projects in Jenkins.")
```
dharmit: `stale_projects` is supposed to be of `list` type. Taking an example:

```python
In [9]: l = [1, 2, 3]

In [10]: if not l:
    ...:     print "list has some entries"
    ...: else:
    ...:     print l
    ...:     
[1, 2, 3]

In [11]: l = []

In [12]: if not l:
    ...:     print "list has some entries"
    ...: else:
    ...:     print l
    ...:     
list has some entries
```

So, it looks to me like this logic is doing exactly opposite of what is expected from it. Am I looking at this properly? Feel free to call it the dumbest question. 😄

navidshaikh: ah well! my bad. fixing it. Thanks for pointing it out!

bamachrn:
LGTM

dharmit:
LGTM, merging.

PR #404, opened 2017-12-19
Adds cron job to delete postgres dump older than N days

navidshaikh:
This changeset adds a cron job to run daily and delete `N` days old pg_dump tar files from NFS directory.
`N` can be configured from `group_vars/all`, default is `N=4`.

### DIFF
--- a/provisions/group_vars/all
+++ b/provisions/group_vars/all
@@ -21,6 +21,8 @@ db_host: '0.0.0.0'
 db_port: 5432
 postgresql_image: registry.centos.org/centos/postgresql-95-centos7
 postgresql_uid: 26
+# expire the cccp database pg_dump tar files after given number of days
+expire_tar_after: 4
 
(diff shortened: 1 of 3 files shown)


provisions/group_vars/all
```diff
@@ -21,6 +21,8 @@ db_host: '0.0.0.0'
 db_port: 5432
 postgresql_image: registry.centos.org/centos/postgresql-95-centos7
 postgresql_uid: 26
+# expire the cccp database pg_dump tar files after given number of days
+expire_tar_after: 4
```
mohammedzee1000: Any reason for this specific value, or is it just something to start off with?

bamachrn:
LGTM

provisions/group_vars/all
navidshaikh: it is something to start off with.

PR #405, opened 2017-12-19
Retries receiving nodes from CICO 3 times while running CI

navidshaikh:
If there is an issue receiving nodes from CICO, retry 3 times, with 1 minute delay.


bamachrn:
LGTM

PR #425, opened 2018-01-10
Fixes initialiation of logger in scan worker runner

navidshaikh:
logger was initialized for scan worker runner before it was referenced
  while trying to connect to docker socket.


navidshaikh:
#dotests

PR #427, opened 2018-01-10
Gates pep8 on service code before running actual tests

navidshaikh:
Gates service code for pep8 coding guidelines to ensure we add code clean and neat.


ci/lib.py
```diff
@@ -525,5 +542,28 @@ def test(data, path=None):
...
+    # setup node and install necessary packages
+    setup_controller(controller)
+
+    # sync controller with service code and patches applied on top
+    sync_controller(controller)
+
+    # CONTROLLER_WORK_DIR constant is the path where source code
+    # on controller is synced at, check constants.py and constant references
+    # run pep8 gate on pipeline service code
+    run_cmd("pep8 %s" % CONTROLLER_WORK_DIR, host=controller, stream=True)
```
dharmit: Can we use the `--exclude` switch to exclude certain files like `beanstalkc.py` and maybe others? I think we need to come up with a list of files that we don't want to lint here.

navidshaikh: @dharmit : done, added a config file where `beanstalkc.py` is excluded.

bamachrn:
#dotests

navidshaikh:
#dotests-debug

dharmit:
So this is going to pep8 and flake8 checks on the code. How does dev know about the results? Are we going to fail the CI check? Is there going to be some message that conveys what pep8/flake8 issues caused failure? Would we have a link or some doc that lists what pep8/flake8 glitches we care more/less about?

navidshaikh:
@dharmit : 
 - This PR blocks CI, if changeset introduces any pep8 issue.
 - Dev can check CI console for issues and its description.
 - in the root directory of repo, find .pep8.config file for behavior.

PR #429, opened 2018-01-10
Fixes pep8 and flake8 issues for python code in scan module

navidshaikh:
Grouped related fixes per module for ease of review.


navidshaikh:
> Can we remove beanstalkc.py from this? It's a third part library. Not written by us.

done

PR #433, opened 2018-01-15
Takes off all the existing migrations and prevent creation of migration part of deployment

navidshaikh:
Since its not a good practice to make migrations on prod environments.


navidshaikh:
@bamachrn @dharmit @mohammedzee1000 : I kept the `migrate` operations as part of ansible scripts, since making migrations should happen prior to deployment however I think we should keep the `migrate` operations as part of ansible itself. This helps eliminate the manual operation errors (say one missed `migrate` step). Also in this patch, we are ensuring `migrate` operation works as expected by removing `ignore_errors` flag.

navidshaikh:
#dotests

navidshaikh:
> I kept the migrate operations as part of ansible scripts, since making migrations should happen prior to deployment however I think we should keep the migrate operation as part of ansible itself.

I tried deployments with mentioned configs however, it could create problems. I tried updating the `migrate` command flags like `python manage.py  --merge` which merges the migrations operations, however it could lead to merge conflicts. I think we should totally, take off the migrations related operations from ansible deployment scripts and have it done manually. Since migrations are not going to be usual operation as compared to usual deployment of service, we can take off the complexity and potential issues that could be involved.

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
The scope of PR is now to take off all the migrations and start afresh. For henceforth, its agreed upon to maintain the migrations as we are going to maintain the data of db.

navidshaikh:
#dotests

navidshaikh:
#dotests

bamachrn:
LGTM, only concern is, we are not loosing data upon re-deployment

PR #443, opened 2018-01-19
Fix unicode issue so that we can process non English characters

dharmit:
This PR fixes the issue that caused failure in building image for https://github.com/CentOS/container-index/pull/256


bamachrn:
@dharmit can we try once failing linter and see notification comes for this? code wise it looks good.

PR #447, opened 2018-02-01
Fixing pattern so that yaml files are also picked up.

mohammedzee1000:
This PR ensures that index ci can pick up on files ending with yaml and not just yml extension.


mohammedzee1000:
@bamachrn @navidshaikh ^

navidshaikh:
@mohammedzee1000 : No description of the PR! PRs should always set context via description field.

mohammedzee1000:
@navidshaikh updated, ptal

bamachrn:
can we think of keeping both, so that the existing ones does not break?

mohammedzee1000:
@bamachrn it should not break charlie's pr (https://github.com/CentOS/container-index/pull/271) updates all files currently in index to yaml.

bamachrn:
I would say we keep both in the check for now. later after everyone moves to yaml, we can go for validating only yaml. otherwise this will cause a huge change across all the repositories we are having in container-index.

mohammedzee1000:
@bamachrn understood. I am not updating the matching for cccp yet. Only the index file we control and those can be updated by us immediately.

PR #449, opened 2018-02-02
Refactoring garbage collector to bring common code to lib.

mohammedzee1000:
This refactor moves common stuff, such as index accessing and registry accessing code and much more into the lib, as its possible to reuse the logic anywhere across the pipeline.


mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

container_pipeline/cleanup_registry/garbagecollector.py
```diff
@@ -136,33 +91,31 @@ def _mark_for_removal(self):
...
-            tags = manifests + "/tags"
-            # Delete the tag
-            for item in tag_list:
-                container_tag = str.format(
-                    "{container_name}/{container_tag}",
-                    container_name=container_full_name,
-                    container_tag=item
+            for tag in tag_list:
+                if "/" in container_full_name:
+                    container_namespace, container_name = container_full_name.split(
```
bamachrn: we are excluding the library images here?

container_pipeline/cleanup_registry/garbagecollector.py
```diff
@@ -181,88 +134,119 @@ def _gcollect(self):
...
-                    else:
-                        if match_only_list_found and not exception_list_found:
-                            self._mismatched[registry_name].append(
-                                registry_tag
-                            )
+        else:
+            registry_info = RegistryInfo(
+                registry_host=self._registry_host,
+                registry_port=self._registry_port,
+                registry_secure=self._registry_secure
```
bamachrn: what registry_secure stand for, https? AFAIK we are doing this locally so how does it matter?

container_pipeline/lib/index.py
```diff
@@ -0,0 +1,82 @@
+import os
+from glob import glob
+from container_pipeline.utils import load_yaml
+
+import container_pipeline.constants.index
+import logging
+import yaml
+
+
+class IndexEntry(object):
```
bamachrn: index reader wont help here? so that we have a common logic everywhere. in case of updating something it gets updated everywhere?

container_pipeline/lib/registry.py
```diff
@@ -0,0 +1,61 @@
...
+            if tags_data:
+                self.tags[item] = json.load(tags_data)
+
+
+def mark_removal_from_local_registry(verbose, container_namespace,
+                                     container_name, container_tag,
+                                     build_running = False):
+    print_msg("Marking mismatched containers for removal...",
+               verbose)
+    registry_storage_path = "/var/lib/registry/docker/registry/v2"
```
bamachrn: may be putting this to some variable would help?

container_pipeline/lib/index.py
mohammedzee1000: Common login should be in lib or utils. That is why i put it there. Also the indexentry thing means you wont be dealing with dict keys and hence less runtime error prone.

container_pipeline/lib/registry.py
mohammedzee1000: +1 will do that :)

container_pipeline/cleanup_registry/garbagecollector.py
mohammedzee1000: This is just to make it future proof. It has no affect now, but doing it now means i wont have to change in case we decide to go full secure or something.

container_pipeline/cleanup_registry/garbagecollector.py
mohammedzee1000: No, we arent. Library images do get cleaned up too.  Tested on my setup.

bamachrn:
@dharmit does this looks good to you for merging? lets take the `except and match` list in next iteration.

PR #451, opened 2018-02-05
Adds docker system prune in cleanup utility

navidshaikh:
This will remove:
	- all stopped containers
	- all volumes not used by at least one container
	- all networks not used by at least one container
	- all dangling images

### DIFF
--- a/provisions/roles/cleanup/scanner/tasks/main.yml
+++ b/provisions/roles/cleanup/scanner/tasks/main.yml
@@ -11,7 +11,12 @@
   sudo: yes
   shell: rm -rf /var/lib/docker/volumes/*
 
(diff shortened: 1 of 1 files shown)


bamachrn:
this looks good. but can we delete the volumes and all after removing the container while scanning them?

navidshaikh:
>  but can we delete the volumes and all after removing the container while scanning them?

We can! I think you missed PR #452.

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

bamachrn:
LGTM, Merging

PR #452, opened 2018-02-05
Prunes dangling images and volumes post scan

navidshaikh:
Added a clean up utility to scan worker systemd service to clean up
  residue post scan.

  Kept the cleanup utility after handle job and not in actual scanner runner
  class since, the clean up can go async with actual scan runner.


navidshaikh:
#dotests

navidshaikh:
#dotests

bamachrn:
LGTM, merging this

PR #455, opened 2018-02-05
Updates the reference to the registry image

cdrage:
Removes the old interface to replace it with the new registry.centos.org
site

### DIFF
--- a/provisions/roles/registry/tasks/main.yml
+++ b/provisions/roles/registry/tasks/main.yml
@@ -16,16 +16,11 @@
 - name: Run UI registry container.
   sudo: yes
   docker_container:
-      image: registry.centos.org/pipeline-images/docker-registry-frontend:v2
+      image: registry.centos.org/pipeline-images/reg:latest
       name: registry-ui
       state: started
       restart: yes
       restart_policy: always
       pull: yes
-      published_ports: "9000:80"
-      env:
-          ENV_DOCKER_REGISTRY_HOST: "{{ public_registry }}"
-          ENV_DOCKER_REGISTRY_PORT: "5000"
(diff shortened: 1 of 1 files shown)


bamachrn:
#dotests

provisions/roles/registry/tasks/main.yml
```diff
@@ -16,16 +16,11 @@
...
       pull: yes
-      published_ports: "9000:80"
-      env:
-          ENV_DOCKER_REGISTRY_HOST: "{{ public_registry }}"
-          ENV_DOCKER_REGISTRY_PORT: "5000"
-          ENV_REGISTRY_PROXY_FQDN: "registry.centos.org"
-          ENV_REGISTRY_PROXY_PORT: "443"
-          ENV_MODE_BROWSE_ONLY: "true"
+      published_ports: "9000:8080"
+      command: "--registry registry.centos.org"
```
bamachrn: how about replacing the `registry.centos.org` with `{{public_registry}}` ? this will allow the registry to be working with deployment other than prod as well.

cdrage: yeah, you're right. updated!

cdrage:
@bamachrn Updated with your changes.

bamachrn:
#dotests

cdrage:
Looks like a false positive. Try again?

mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

mohammedzee1000:
#dotests

bamachrn:
I believe it needs manual try after we got the current master deployed on prod. running tests does not make much sense. But ideally it should have cleared CI as nothing getting broken here.

cdrage:
Tests all green 💯 anything else for me to do my side?

PR #456, opened 2018-02-06
Updates scanners functional tests to expect absence of execution status files

navidshaikh:
Mail server cleans execution status files post-processing, the tests are updated to expect absence of status file.

Also updates misc package updates scanner test, since it refuses to generate result if underlying tool used for scanning is absent.

### DIFF
--- a/ci/tests/test_01_functional/test_03_scanner/__init__.py
+++ b/ci/tests/test_01_functional/test_03_scanner/__init__.py
@@ -5,8 +5,7 @@ import time
 
 from random import randint
 from ci.tests.base import BaseTestCase
-from container_pipeline.lib.settings import SCANNERS_RESULTFILE,\
-        SCANNERS_STATUS_FILE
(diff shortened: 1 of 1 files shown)


navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

navidshaikh:
#dotests

bamachrn:
In this case we are saying we run the tests after mail is sent. did not get how this was required?

navidshaikh:
> In this case we are saying we run the tests after mail is sent. did not get how this was required?

If you check the tests, there is `time.sleep` at intervals and checking for scanners results/execution status files. Also mail server removes the execution status file. So there is a possibility that, between the sleep period, scanners just finished and mail is sent, thus no presence of execution status file. Thats why the tests are updated.

bamachrn:
understood, Lets get this in.

PR #473, opened 2018-02-12
Don't verify TLS certificates for registry.

cdrage:
Sometimes the TLS certs are self-generated when deploying. This ignores
invalid certs.

### DIFF
--- a/provisions/roles/registry/tasks/main.yml
+++ b/provisions/roles/registry/tasks/main.yml
@@ -23,4 +23,4 @@
       restart_policy: always
       pull: yes
       published_ports: "9000:8080"
-      command: "--registry {{ public_registry }}"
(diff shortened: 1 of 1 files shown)


cdrage:
ping @navidshaikh

bamachrn:
#dotests

navidshaikh:
Should be rephrased to "dont verify TLS for all the dev/test environments" :wink: .

cdrage:
I'll update the PR in a bit to include differences between production / dev environments considering we have a deployment=production key as pointed out by @navidshaikh

mohammedzee1000:
@cdrage are we good with the changes here?

cdrage:
@mohammedzee1000 Yup. Good with it for now. Let's implement the production / dev scripting later.

PR #476, opened 2018-02-13
[WIP]Avoid searching for target file in the pre-build request entries

bamachrn:
* this PR fixes most of the pep8 issues in `container-index validator.py`
* skips the pre-build request entries to not search for `dockerfile`

### DIFF
--- a/testindex/validators.py
+++ b/testindex/validators.py
@@ -240,9 +240,7 @@ class IndexFormatValidator(IndexValidator):
                 depends_on = entry["depends-on"]
                 if not isinstance(depends_on, list):
                     depends_on = [depends_on]
-                matcher = re.compile("^(([0-9a-zA-Z_-]+[.]{1})*([0-9a-zA-Z_-]+)\
-                                     {1}[/]{1})?[0-9a-zA-Z_-]+[/]{1} \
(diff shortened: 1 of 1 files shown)


testindex/validators.py
```diff
@@ -240,9 +240,7 @@ def run(self):
                 depends_on = entry["depends-on"]
                 if not isinstance(depends_on, list):
                     depends_on = [depends_on]
-                matcher = re.compile("^(([0-9a-zA-Z_-]+[.]{1})*([0-9a-zA-Z_-]+)\
-                                     {1}[/]{1})?[0-9a-zA-Z_-]+[/]{1} \
-                                     [0-9a-zA-Z_-]+[:]{1}([0-9a-zA-Z_-]+\.?)+$"
+                matcher = re.compile("^(([0-9a-zA-Z_-]+[.]{1})*([0-9a-zA-Z_-]+){1}[/]{1})?[0-9a-zA-Z_-]+[/]{1}[0-9a-zA-Z_-]+[:]{1}([0-9a-zA-Z_-]+\.?)+$"
```
navidshaikh: I would suggest, we add some tests with different samples to be tested with regex.

PR #477, opened 2018-02-13
Ignore centos base images from repo tracking

dharmit:
This PR is meant to ignore images starting with `centos/centos` from repo tracking (RPM updates).


dharmit:
#dotests

PR #486, opened 2018-02-22
Fixes bug in dangling volume cleanup post scan

navidshaikh:
This changeset fixes invoking the dangling volume clean up command.

### DIFF
--- a/container_pipeline/workers/scan.py
+++ b/container_pipeline/workers/scan.py
@@ -108,7 +108,7 @@ class ScanWorker(BaseWorker):
         command = "docker rmi -f `docker images -f dangling=true -q`"
         self.logger.debug("Removing unused/dangling images..")
         try:
-            run_cmd(command)
+            run_cmd(command, shell=True)
         except Exception as e:
             self.logger.critical("Failing to remove dangling images.")
             self.logger.critical("Error: %s", str(e))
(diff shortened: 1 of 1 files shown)


bamachrn:
did we have a test run for this? I have a doubt on output logs.

dharmit:
I'd like us to test this PR by doing a clean deployment based off current master and use an index that's got about 10 or so images. More the images, less the manual overhead of triggering builds. So one of the images could be a base image that other images are dependent on.

Next, let it scan the images and let's trigger a few weekly scans on that environment. All this to ensure that we're doing a lot of scanning.

After a few iterations, we need to check if dangling volumes are building up on the scanner node:

```bash
$ docker volume ls -f dangling=true
```

Dangling volumes on scanner node are a confirmation of the problem. Issue is, the problem we're trying to solve here is not getting reproduced properly in pre-prod and devcloud.

With that in place, we can do a deployment of this PR (without cleaning anything.) If everything's working as expected, we should see dangling volumes getting cleaned up.

WDYT @navidshaikh @bamachrn ?

bamachrn:
Yes I was just confirming, @navidshaikh is doing the testing as he said. Or else I can do the testing of this PR.

navidshaikh:
On a fresh deployment to test this PR, projects built and weekly scanning completed for:

```
{
"repositories": [
"centos/centos7-atomic",
"centos/mariadb-101-centos7",
"centos/nodejs-4-centos7",
"centos/nodejs-6-centos7",
"centos/nodejs-8-centos7",
"centos/passenger-40-centos7",
"centos/perl-520-centos7",
"centos/perl-524-centos7",
"centos/ruby-22-centos7",
"centos/ruby-23-centos7",
"centos/ruby-24-centos7",
"centos/s2i-core-centos7",
"nshaikh/go-helloworld"
]
}
```

The volume listing on scanner node post builds and weekly scanning:
```
[root@scanner ~]# docker volume ls -f dangling=true
DRIVER              VOLUME NAME
```
@bamachrn @dharmit : I think this is good to go.

PR #488, opened 2018-02-26
Ensure raven is installed

cdrage:
Ensures that Raven is installed when deploying Django.

### DIFF
--- a/provisions/roles/django/tasks/main.yml
+++ b/provisions/roles/django/tasks/main.yml
@@ -34,6 +34,11 @@
   sudo: yes
   tags: django
 
+- name: Install raven
+  pip: name=raven version=6.5.0
+  sudo: yes
+  tags: django
+
 - name: Ensure docker group exists
   group: name=docker state=present
   sudo: yes


provisions/roles/django/tasks/main.yml
```diff
@@ -34,6 +34,11 @@
   sudo: yes
   tags: django
 
+- name: Install raven
+  pip: name=raven version=6.5.0
```
bamachrn: is there any other way than `pip` to get raven? generally `pip` is not preferred to take to production

cdrage: @bamachrn We install Django as well through Pip...

I can't find where, but usually 6.5.0 of raven is installed (I believe through Django? Or by default within CentOS7 regular OS not container). 

The reason I added this is because I was getting errors of raven not being installed when deploying CCCP to a container-cluster.

Looks like an old version is being used within CentOS, so preferably, it may be better to install via `pip`, see: https://centos.pkgs.org/7/epel-x86_64/python-raven-5.12.0-1.el7.noarch.rpm.html

bamachrn:
LGTM

PR #490, opened 2018-02-26
Force creation of cron.hourly directory

cdrage:
Some CentOS distributions do not have cron installed by default (see
CentOS7 Docker container).

Thus force=yes is required to manually generate / create the folder.


bamachrn:
LGTM :+1:

PR #491, opened 2018-02-26
Improves deployment of Jenkins via Ansible

cdrage:
This PR:

 - Stops the Jenkins master (via systemd) BEFORE configuration takes
 place. There are minor issues where if the configuration is overriden
 while Jenkins is running, configuration settings are potentially
 overridden. After configuration changes are completed, Jenkins is
 started again.
  - Uses ansible's built-in `wait_for` for port 8080 to come up, then
  checks to see if the UI has come up. This reduces the delay duration
  as well as makes it more intuitive (wait for port to come up, then
  curl it)
  - Reduces delays (faster deployment)
  - Makes sure that a shell is created for Jenkins slave user (had
  issues with master+slave communication due to user being directed to
  /run/nologin for some reason)

### DIFF
--- a/provisions/roles/jenkins/master/tasks/main.yml
+++ b/provisions/roles/jenkins/master/tasks/main.yml
@@ -2,22 +2,39 @@
   tags:
   - jenkins/master
 
+# Stop Jenkins so no files are overriden by a running Jenkins instance
+- name: Stop jenkins before configuration
+  service: name=jenkins state=stopped
+  sudo: yes
+  tags:
+  - jenkins/master
+
 - include: config.yml
   tags:
   - jenkins/master
 
-- name: Restart jenkins
-  service: name=jenkins state=restarted
+# Start Jenkins again
+- name: Start jenkins
+  service: name=jenkins state=started
   sudo: yes
   tags:
   - jenkins/master
 
 - name: Wait for Jenkins to come up
-  shell: curl --head --silent http://localhost:8080/cli/
+  wait_for:
+    host: '{{ansible_default_ipv4.address}}'
+    port: 8080
+    delay: 5
+    timeout: 300
+  tags:
+    - jenkins/master
+
+- name: Wait for Jenkins UI to come up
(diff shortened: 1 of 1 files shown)


provisions/roles/jenkins/master/tasks/main.yml
```diff
@@ -2,22 +2,39 @@
   tags:
   - jenkins/master
 
+# Stop Jenkins so no files are overriden by a running Jenkins instance
+- name: Stop jenkins before configuration
```
bamachrn: this was being taken care of by the cleanup task though.

provisions/roles/jenkins/master/tasks/main.yml
```diff
@@ -2,22 +2,39 @@
...
+# Start Jenkins again
+- name: Start jenkins
+  service: name=jenkins state=started
   sudo: yes
   tags:
   - jenkins/master
 
 - name: Wait for Jenkins to come up
-  shell: curl --head --silent http://localhost:8080/cli/
+  wait_for:
```
bamachrn: thanks for bringing this in, this will save us a lot of time

provisions/roles/jenkins/master/tasks/main.yml
```diff
@@ -31,11 +48,20 @@
...
+    host: '{{ ansible_default_ipv4.address }}'
+    port: 8080
+    delay: 5
+    timeout: 300
+  tags:
+    - jenkins/master
+
+- name: Wait for Jenkins UI to come up
+  shell: curl --head --silent http://{{ ansible_default_ipv4.address }}:8080/cli/
```
bamachrn: :+1: :100:

provisions/roles/jenkins/master/tasks/main.yml
cdrage: Yeah.. Problem I was facing was that configurations were *still* being overriden unless I had this added in.

PR #492, opened 2018-02-26
Allow ability to run Docker-in-Docker-in-OpenShift for testing

cdrage:
A bit of a complicated PR, but essentially, after much trial-and-error,
this is required in order to get OpenShift running within
Docker-in-Docker within a CentOS7 container.

Essentially, what happens, is that OpenShift tries to start Kubelet but
is unable to due to cgroup access within the CentOS7 Container host.

See issue: https://github.com/kubernetes/kubernetes/issues/43856

In the end, it is instead required to:

 1. Generate the default configuration for the all-in-one OpenShift
 origin container
 2. Edit node-config.yaml to ignore QoS containers
 3. Start the OpenShift origin container.

This will only occur if the `deployment` value is set to `test`

Another note is that `selinux` has been added as a tag to "Set SELinux
context for openshift shared dirs" due to selinux not being deployed /
used on CentOS7 containers. This is added so we can use --skip-tag
"selinux" if we are deciding to deploy to CentOS7 containers.

### DIFF
--- a/provisions/roles/openshift/defaults/main.yml
+++ b/provisions/roles/openshift/defaults/main.yml
@@ -17,7 +17,7 @@ openshift_shared_dirs:
 openshift_volumes:
   - /:/rootfs:ro
   - /var/run:/var/run:rw
-  - /sys:/sys
+  - /sys:/sys:rw
   - /sys/fs/cgroup:/sys/fs/cgroup:rw
   - /var/lib/docker:/var/lib/docker:rw
   - /var/lib/origin/openshift.local.volumes:/var/lib/origin/openshift.local.volumes:rslave
--- a/provisions/roles/openshift/tasks/config.yml
+++ b/provisions/roles/openshift/tasks/config.yml
@@ -19,6 +19,7 @@
 - name: Set SELinux context for openshift shared dirs
   command: chcon -Rt svirt_sandbox_file_t {{ item }}
   with_items: "{{ openshift_shared_dirs }}"
+  tags: selinux
 
 - name: Pull openshift images
   docker_image:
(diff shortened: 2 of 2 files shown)


cdrage:
#runtests

cdrage:
I believe this is a false-positive as I do not touch Jenkins at all.

bamachrn:
#dotests

bamachrn:
I am a bit sceptical about running docker-in-docker. Not getting why we are putting efforts to get openshift running within container.

cdrage:
@bamachrn 

Three reasons!

1. 

This is for testing in order to try to replace the Vagrant suite that we currently have. Running this within Virtualbox / KVM ended up taking over an hour just to deploy on a laptop. 

With containers, I was able to deploy this within 20 minutes (sometimes 15 if I close Chrome / all other applications). On a desktop (Core i5, 12GB RAM), deployment was less than 10. 

2.

Another reason is making the Ansible playbooks compatible with https://github.com/ansible/ansible-container 

Instead of using Dockerfile's, we could use what we already have (Ansible playbooks) and combine it with https://github.com/ansible/ansible-container in order to great OpenShift / Kubernetes deployable containers.

Of course, we'd modify the Ansible playbooks further in the future (would be redundant to have OpenShift run OpenShift, haha). But this is the first step towards it :+1: 

3.

Thirdly, this makes CCCP compatible with alternative testing suites where virtualization is too much overhead or incompatible. Having the ability to deploy CCCP to containers rather than virtual machines allows the ability to run a CCCP cluster on SemaphoreCI / Travis / etc with much lower overhead.

cdrage:
#dotests

cdrage:
#dotests

cdrage:
@bamachrn @navidshaikh @mohammedzee1000 One of you mind doing a PR review? This is my last PR before I'm unblocked from pushing my `dev.sh` script / development documentation.

provisions/roles/openshift/defaults/main.yml
```diff
@@ -17,7 +17,7 @@ openshift_shared_dirs:
 openshift_volumes:
   - /:/rootfs:ro
   - /var/run:/var/run:rw
-  - /sys:/sys
+  - /sys:/sys:rw
```
navidshaikh: should this change be applicable only when `deployment==docker`?

provisions/roles/openshift/tasks/config.yml
```diff
@@ -31,7 +32,56 @@
...
+    image: "{{ origin_image_registry }}/{{ origin_image_name }}:{{ origin_image_tag }}"
+    privileged: yes
+    pid_mode: host
+    network_mode: host
+    recreate: yes
+    volumes: "{{ openshift_volumes }}"
+    command: start --master=https://"{{ansible_default_ipv4.address}}":8443 --write-config=/var/lib/origin/openshift.local.config
+
+- name: Quick pause to catch up to configuration being propagated
+  when: deployment == "test"
```
navidshaikh: I think `deployment == "docker"` here?

provisions/roles/openshift/tasks/config.yml
```diff
@@ -31,7 +32,56 @@
...
+    - 'kubeletArguments:'
+    - '  cgroups-per-qos:'
+    - '  - "false"' 
+    - '  enforce-node-allocatable:'
+    - '  - ""'
+
+- name: Create test OpenShift origin container
+  when: deployment == "docker"
+  docker_container:
+    name: origin
```
navidshaikh: should the name clash with actual "production OpenShift origin container" ([immediate next task](https://github.com/CentOS/container-pipeline-service/pull/492/files#diff-8b0f8f52a368116ae8035924236a6cb9R86))?

How about `name: origin_test`?

provisions/roles/openshift/defaults/main.yml
cdrage: this actually doesn't matter as by default it's read/write for volumes. the only reason why I put `:rw` is to make it consistent with the rest of the volumes.

cdrage:
@navidshaikh Updated it based on your changes, thanks!

cdrage:
@navidshaikh that'll come next in my developer doc / script!

PR #494, opened 2018-02-27
Installs latest docker and docker-py packages on every deployment

### LINKED ISSUE #493: Inconsistent docker package version on service nodes
With a couple of nodes (slave, OpenShift) installing latest docker package with `state=latest` while dependencies for rest node are satisfied with `state=present` config in ansible provisioning, leaving nodes with inconsistent docker package version during re-deployments.

navidshaikh:
Fixes #493

  Updated the ansible yum install config of `docker` and `python-docker-py`
  packages to `state=latest`.
  Earlier the config was `state=present`.
  This was causing inconsistent version of docker installed on different
  nodes.

### DIFF
--- a/provisions/roles/django/tasks/main.yml
+++ b/provisions/roles/django/tasks/main.yml
@@ -14,14 +14,12 @@
   sudo: yes
   tags: django
 
-- name: Install Docker
-  yum: name=docker state=installed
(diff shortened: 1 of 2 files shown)


bamachrn:
LGTM, merging this

PR #499, opened 2018-03-01
Remove postgresql container that's not being used anywhere

dharmit:
Stop pulling registry.centos.org/sclo/postgresql-95-centos7 container as it's not being used anywhere in the service.

### DIFF
--- a/provisions/roles/db/tasks/main.yml
+++ b/provisions/roles/db/tasks/main.yml
@@ -37,13 +37,6 @@
   sudo: yes
   systemd: name=docker state=started enabled=yes
 
-- name: Pull CentOS postgres container
-  docker_image:
-      name: registry.centos.org/sclo/postgresql-95-centos7
-      tag: latest
-  sudo: yes
-  tags: db
-
 - name: Run postgres container
   docker_container:
       name: postgres
--- a/provisions/roles/db/tasks/main.yml
+++ b/provisions/roles/db/tasks/main.yml
@@ -41,6 +41,7 @@
   docker_image:
       name: '{{ postgresql_image }}'
       tag: latest
+      state: present
       force: yes
   sudo: yes
   tags: db
--- a/provisions/roles/db/tasks/main.yml
+++ b/provisions/roles/db/tasks/main.yml
@@ -37,6 +37,14 @@
   sudo: yes
   systemd: name=docker state=started enabled=yes
 
+- name: Pull CentOS postgres container
+  docker_image:
+      name: '{{ postgresql_image }}'
+      tag: latest
+      force: yes
+  sudo: yes
+  tags: db
+
 - name: Run postgres container
   docker_container:
       name: postgres


dharmit:
This is more of a test PR right now till the CI is green. In our codebase, we have set:

```postgresql_image: registry.centos.org/centos/postgresql-95-centos7```

for use of DB that's running on Jenkins master. However, we're still pulling similar container from SCLo namespace. Also, most SCLo containers moved to CentOS namespace via https://github.com/CentOS/container-index/pull/219/

**Comment by dharmit** [comment]
#dotests

**Comment by dharmit** [comment]
#dotests

File: provisions/roles/db/tasks/main.yml [hunk 1]
```diff
@@ -37,13 +37,6 @@
   sudo: yes
   systemd: name=docker state=started enabled=yes
 
-- name: Pull CentOS postgres container
-  docker_image:
-      name: registry.centos.org/sclo/postgresql-95-centos7
```
navidshaikh: I would suggest we replace line no 42, the hardcoded name of `sclo` namespaced image with `'{{ postgresql_image }}'`.
Thus, there is `pull` and `run` tasks.

dharmit: Agree. But we don't need to explicitly do it since that's taken care of [here](https://github.com/CentOS/container-pipeline-service/blob/22dd7b52edaa0f5ab9e2fba8c89c262ce9e35b67/provisions/roles/db/tasks/main.yml#L47-L62). `docker_container` module pulls an image if one it's supposed to create container out of doesn't exist.

navidshaikh: I know that. I wanted to convey about dividing pull and run between two tasks for better control. Like if you can update the pull image task with `force: yes`, it will ensure that always there is an effort to pull (which will ensure during every deployment image is upto date) before run.

dharmit: I had same concern for using an up-to-date image during each deployment. But that warrants a different PR, right? wdyt?

navidshaikh: I think we can just remove the hardcoded `sclo` repository path and use the variable. Add another commit to same PR, saying add force pull image on every deployment, update PR title, we are good!

dharmit: sounds good. made changes. PTAL.

provisions/roles/db/tasks/main.yml
```diff
@@ -39,8 +39,9 @@
 
 - name: Pull CentOS postgres container
   docker_image:
-      name: registry.centos.org/sclo/postgresql-95-centos7
+      name: '{{ postgresql_image }}'
       tag: latest
+      force: yes
```
navidshaikh: I think we'll need `state: present` as well @dharmit , since `force` functionality is clubbed with `state` config.
PTAL at [this](http://docs.ansible.com/ansible/latest/docker_image_module.html) doc, the force table entry.

dharmit:
#dotests

navidshaikh:
#dotests

PR #502, opened 2018-03-02
Removes common variables between ansible group_var and inventory file

### LINKED ISSUE #466: Cleanup variables + actual hosts file / problems with setting multiple variables within the hosts file.
Okay... So variables from `group_vars/all` is overriding ANY variables which are being set within the hosts file..

For example, if I set:

```
test=True
```

Within my `hosts.sample` file. It will *never* propagate as `True` since `group_vars/all` is set to False.

All variables from within `group_vars/all` should instead be located in the example `hosts.sample` file rather than being located in a separate folder.

navidshaikh:
Fixes #466
  - Re-organized variables in inventory file
  - Moved database related variables in group_vars/all file

### DIFF
--- a/provisions/group_vars/all
+++ b/provisions/group_vars/all
@@ -1,19 +1,17 @@
 ---
+
+# vagrant is no more supported, keep it False
 vagrant: False
-rsync_ssh_opts: ""
-test: false
-setup_nfs: false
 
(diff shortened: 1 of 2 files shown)


cdrage:
LGTM

bamachrn:
LGTM

PR #503, opened 2018-03-02
Fixes atomic scanner installation during provisioning

navidshaikh:
Instead of using docker run command, uses atomic install command.


navidshaikh:
#dotests

navidshaikh:
#dotests

PR #506, opened 2018-03-06
Moves db_host variable from group_vars to inventory file

navidshaikh:
Since db_host needs a jenkins-master IP address, and needs changes per deployment.

### DIFF
--- a/provisions/hosts.sample
+++ b/provisions/hosts.sample
@@ -33,6 +33,7 @@ public_registry=jenkins-slave
 intranet_registry=jenkins-slave:5000
 # configure beanstalkd server node
 beanstalk_server=openshift
+db_host=jeankins-master
 
 
 ## Container Index related config
--- a/provisions/hosts.sample
+++ b/provisions/hosts.sample
@@ -11,10 +11,10 @@ jenkins-master
 jenkins-slave
 
 [openshift]
-openshift
+open-shift
 
 [scanner_worker]
-scanner_worker
+scanner-worker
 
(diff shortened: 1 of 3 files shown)


provisions/hosts.sample
```diff
@@ -32,7 +32,8 @@ log_level = DEBUG
 public_registry=jenkins-slave
 intranet_registry=jenkins-slave:5000
 # configure beanstalkd server node
-beanstalk_server=openshift
+beanstalk_server=open-shift
+db_host=jeankins-master
```
cdrage: jeankins?

provisions/hosts.sample
```diff
@@ -11,10 +11,10 @@ jenkins-master
 jenkins-slave
 
 [openshift]
-openshift
+open-shift
```
cdrage: why is it like `open-shift`

provisions/hosts.sample
navidshaikh: ah, that's a typo, fix on the way!

provisions/hosts.sample
navidshaikh: This is to match the regex and avoid using the name in square brackets `[openshift]`. This will help in vim/editors to replace all the occurrences of `open-shift` with its IP/FQDN, like
```
:%s/open-shift/192.168.1.3/g
```
if I keep it `openshift` and run the regex, the undesired will be replaced too, for eg
```
[192.168.1.3]
192.168.1.3
```
Make sense ?

PR #512, opened 2018-03-07
Updating database to include link to target file and removing old method

mohammedzee1000:
This PR removes the old save in file method, in favour of saving link to target file in the DB itself.

This will be the last database PR without maintained migrations. From here on out, any changes to the database will compulsorily require migrations.


navidshaikh:
@mohammedzee1000 : Commit the in-effect migrations to this PR. You should be able to generate it against a running service in devcloud.

mohammedzee1000:
@navidshaikh technically yes, but we are going for clean db and that has not been deployed yet so we should be good in that regard. Once the first deployment happens with clean db, we will have to ensure the migrations are maintained.

mohammedzee1000:
@navidshaikh changes done
@bamachrn Review please

navidshaikh:
CI console has errors needing fix.
```
./container_pipeline/utils.py:239:20: W292 no newline at end of file
./jenkinsbuilder/cccp_index_reader.py:130:25: E303 too many blank lines (2)
Build failed as pep8 checks failed tests failed.
```

jenkinsbuilder/weekly-scan.py
```diff
@@ -111,7 +116,15 @@
             job = bs.put(json.dumps(data))
 
             # Initializing Database entries
-            project, created = Project.objects.get_or_create(name=project_name)
+            project, created = Project.objects.get_or_create(
+                name=project_name,
+                target_file_link=form_targetfile_link(
+                    git_url,
```
navidshaikh: why add an extra variable? Just use `entry["git-url"]`

same applies to following 3 more variables used.

mohammedzee1000: That is in keeping with rest of the pack. In that sense, there is no need for appid, jobid and desired tag either. It simply reduces the possibility of errors as intelli sense, and for that matter, compiler, can handle variable errors, but not mismatched strings.

navidshaikh: Well, the inefficient practices followed in the pack do not really need to be followed for the sake of similarity, while I don't expect the other part of the code to be also fixed with this PR. To comment on the compiler and variable errors, well, the tests should be there to take care of it.

You can keep it as is.

container_pipeline/utils.py
```diff
@@ -211,3 +211,30 @@ def complete(self):
...
+        link_url = str.format(
+            "{git_url}/blob/{git_branch}/{git_path}/{target_file}",
+            git_url=git_URL,
+            git_branch=git_branch,
+            git_path=git_path if not git_path.startswith("/") else "/" +
+                                                                   git_path,
+            target_file=target_file
+        )
+    return link_url
+
```
navidshaikh: The pep8 error read now new line at the end of file, while we are adding an extra line here, causing CI failure at pep8 gate again.

You can run `pep8 utils.py` before doing git push.
Running `autopep8 -i utils.py` should fix utils.py inline.

container_pipeline/utils.py
```diff
@@ -211,3 +211,29 @@ def complete(self):
...
+    Helper function to generate Dockerfile link.
+    :param git_URL: The url of the git repository.
+    :param git_path: The path, relative to the git repository root,
+                     where the file resides.
+    :param git_branch: The repository branch where the file resides.
+    :param target_file: The name of the target file.
+    :return: The reachable link to the dockerfile.
+    """
+
+    # TODO : Move this into a lib.
```
dharmit: If this is a new function being added and since the `TODO` suggests moving it to a different file, what's blocking us from doing it upfront?

container_pipeline/utils.py
```diff
@@ -211,3 +211,29 @@ def complete(self):
...
+
+    link_url = None
+    if "github" in git_URL or "gitlab" in git_URL:
```
dharmit: What happens when someone's not using either GitHub or GitLab? What if they're using BitBucket instead? We return `None` as the value of `link_url`, right? Is that valid for the pieces that make use of this function?

ci/tests/test_01_functional/test_02_linter/__init__.py
```diff
@@ -62,8 +62,18 @@ def start_build(self):
...
+            'name=\\"nshaikh-build-fail-test-latest\\",'
+            'target_file_link=form_targetfile_link('
+            '{}, {}, {}, {}'
+            ')'
+            ')'.format(
+                self.repo_url,
+                self.repo_build_path,
+                self.repo_branch,
+                self.target_file
+            )
```
navidshaikh: simplify this like
```
target_file_link = form_targetfile_link(
    self.repo_url, self.repo_build_path, self.repo_branch, self.target_file)

self.run_dj_script(
'from container_pipeline.models import Project; '
'Project.objects.get_or_create('
'name=\\"nshaikh-build-fail-test-latest\\",'
'target_file_link={}'.format(target_file_link)
```

container_pipeline/models/pipeline.py
```diff
@@ -23,7 +23,7 @@ class Project(models.Model):
     service
     """
     name = models.CharField(max_length=200, unique=True, db_index=True)
-
+    target_file_link = models.CharField(max_length=350, default=None)
```
navidshaikh: We should not have this field in `Projects` table.
The right place for this field is in `Build` table. As `target_file_link` can be changed per build, and every field should reflect correct link for target file.

If you are concern retrieving the link given a project-name, the build table has ForeginKey to Projects table, and you can retrieve builds of given project in descending order to grab the link from latest build.

container_pipeline/utils.py
mohammedzee1000: Yes, i need to remove that TODO.

container_pipeline/utils.py
mohammedzee1000: That is something i need to think about, going forward,.

container_pipeline/models/pipeline.py
mohammedzee1000: Well, i would vote for that too, but there is one slight issue. The link might get invalidated, if the user does an update to, say the git-path. 

The only way i can think of getting around this right now, is to store a copy of the source dockerfile on our end, and im not sure we want to do that.

container_pipeline/utils.py
dharmit: You mean not to be worked upon in this PR?

container_pipeline/models/pipeline.py
navidshaikh: > The link might get invalidated

I think that's fine. It gives the record of what Dockerfile (path) being used to build for a particular build. We should not maintain the Dockerfiles themselves, rather we need to maintain and record artifacts we used for builds.

cdrage:
I saw your ping. I see how you're changing this, but how are you going to make this publicly accessible? Ideally having everything through the API would be the best place.

mohammedzee1000:
@cdrage it will be available through the api, but i was asking for per build vs per project.

cdrage:
@mohammedzee1000 Per project / container in my opinion.

There needs to be a very simple way of just being able to `curl` say: `cdrage/mosh` from the api and get *all* information (last build, dockerfile, etc.) from one simple end-point without having to parse / iterate through everything.

navidshaikh:
#dotests

container_pipeline/models/pipeline.py
```diff
@@ -23,7 +23,7 @@ class Project(models.Model):
     service
     """
     name = models.CharField(max_length=200, unique=True, db_index=True)
-
+    target_file_link = models.CharField(max_length=350, default=None, blank=True, null=True)
```
dharmit: Can you please make sure that this line is less than 80 characters long? It's causing the [CI to fail](https://ci.centos.org/job/centos-container-pipeline-service-ci-pr/1970/console).

container_pipeline/utils.py
mohammedzee1000: yes, but i will update it as soon as i can

mohammedzee1000:
#dotests

mohammedzee1000:
wow green ci, finally

navidshaikh:
@mohammedzee1000 : Since this is updating the db schema, what's your plan for migrations?

mohammedzee1000:
@navidshaikh we were going for a new DB right?

navidshaikh:
> @navidshaikh we were going for a new DB right?

I meant the same, if you can write down the plan in PR description itself?

mohammedzee1000:
@navidshaikh updated.

bamachrn:
LGTM, except the migrations bit. merging it

PR #517, opened 2018-03-12
No need of restart retry when restart is set to always

bamachrn:
When container is set to always restart, docker complains on seeing the restart retries. This PR removes the restart retries to make the container work on restart always mode.


provisions/roles/openshift/tasks/config.yml
```diff
@@ -41,7 +41,6 @@
     volumes: "{{ openshift_volumes }}"
     command: start --master=https://"{{ansible_default_ipv4.address}}":8443
     restart_policy: always
-    restart_retries: 2
```
mohammedzee1000: Dunno what this was for?

Issue #515, opened 2018-03-09
Upgrade gc to delete image revisions in registry and add pagination


mohammedzee1000:
The GC needs to be updated to 

1. Add pagination, so that dynamic growth of registry does not affect how many images are removed
2. Delete image revisions. Basically, all tagged latest images in the registry receive 2 tags, one by named tag, and another by a sha256 manifest. Currently, only tags are cleared, hence the manifest ref to said layers remains.

bamachrn:
PR merged

PR #518, opened 2018-03-12
Updating GC with API pagination and revision tags.

mohammedzee1000:
https://github.com/CentOS/container-pipeline-service/issues/515

- [x] Adding pagination for querying registry API for catalog.
- [x] Adding ability to clean up revision tags.

### DIFF
--- a/container_pipeline/lib/shell/registry_delete_revision_tags.sh
+++ b/container_pipeline/lib/shell/registry_delete_revision_tags.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+REG_PATH="/var/lib/registry/docker/registry/v2/repositories/";
+MANIFEST_LIST="/tmp/registry_manifests.txt"
+
+pushd ${REG_PATH}
+
+echo "Searching for revision tags to remove ... ";
+
+MANIFESTS_WITHOUT_TAGS=$(comm -23 <(find . -type f -name "link" | grep "_manifests/revisions/sha256" | grep -v "\/signatures\/sha256\/" | awk -F/ '{print $(NF-1)}' | sort) <(for f in $(find . -type f -name "link" | grep "_manifests/tags/.*/current/link"); do cat ${f} | sed 's/^sha256://g'; echo; done | sort))
+find . > ${MANIFEST_LIST};
+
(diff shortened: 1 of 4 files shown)


container_pipeline/lib/shell/registry_delete_revision_tags.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+REG_PATH="/var/lib/registry/docker/registry/v2/repositories/";
+MANIFEST_LIST="/tmp/registry_manifests.txt"
+
+pushd ${REG_PATH}
+
+echo "Searching for revision tags to remove ... ";
+
+MANIFESTS_WITHOUT_TAGS=$(comm -23 <(find . -type f -name "link" | grep "_manifests/revisions/sha256" | grep -v "\/signatures\/sha256\/" | awk -F/ '{print $(NF-1)}' | sort) <(for f in $(find . -type f -name "link" | grep "_manifests/tags/.*/current/link"); do cat ${f} | sed 's/^sha256://g'; echo; done | sort))
```
navidshaikh: should you simplify this operation? or decorate a complex operation with comments about whats being done here?

container_pipeline/utils.py
```diff
@@ -1,12 +1,12 @@
...
 import json
-import yaml
 import logging
 import os
-import urllib2
 import subprocess
+import urllib2
 from shutil import rmtree
 
+import yaml
```
navidshaikh: you can group direct `import` and `from ` imports

container_pipeline/utils.py
```diff
@@ -33,27 +33,45 @@ def load_yaml(yaml_file):
     return data
 
 
-def run_cmd(cmd, check_call=True, no_shell=False):
+def run_cmd(cmd, check_call=True, no_shell=False, use_shell=False,
```
navidshaikh: why repeat the utilities code? check `container-pipeline-service/container_pipeline/lib/command.py` add necessary options in that utility function with proper defaults (to keep up with other code already using it)

mohammedzee1000: That is something we need to do, consolidate into one lib. We have various run_cmd and other such funcs scattered accross the board, which i too dont like. Maybe a seperate PR is needed.

PR #519, opened 2018-03-13
Include yaml and yaml both and remove catalog check from weekly scan

bamachrn:
* This PR includes all entries of yaml and yml both from container-index for weekly scan check
* Stops verifying the image to be existing before triggering the scan. This assumes that scanners skips the images tagged as null.

### DIFF
--- a/jenkinsbuilder/weekly-scan.py
+++ b/jenkinsbuilder/weekly-scan.py
@@ -8,7 +8,7 @@ on the registry and initializing the scan tasks for the workers.
 """
 
 import beanstalkc
-import container_pipeline.lib.dj
+# import container_pipeline.lib.dj
 from container_pipeline.models.pipeline import Project, Build, BuildPhase
 from django.utils import timezone
 import glob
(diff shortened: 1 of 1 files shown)


bamachrn:
ping @dharmit @navidshaikh we need to get this in before weekend to have weekly-scan running properly

jenkinsbuilder/weekly-scan.py
```diff
@@ -8,7 +8,7 @@
 """
 
 import beanstalkc
-import container_pipeline.lib.dj
+# import container_pipeline.lib.dj
```
navidshaikh: commenting this out won't configure the django for weekly scan, please uncomment this line to load all the modules, check the dj script
```
$ cat container_pipeline/lib/dj.py 
import django
import os


def load():
    os.environ.setdefault('DJANGO_SETTINGS_MODULE',
                          'container_pipeline.lib.settings')
    django.setup()

load()
```

bamachrn: we are importing this but not calling the load function anywhere. ok uncommenting it.

navidshaikh: @bamachrn 

> we are importing this but not calling the load function anywhere.

I don't think you will need to call the `load()` function explicitly. You can see, the load function is explicitly called in `dj.py`, so even if you import the dj script, the `load` will be called at the start of the file.

bamachrn:
This is working properly now. would you please approve and merge this.

jenkinsbuilder/weekly-scan.py
```diff
@@ -34,21 +34,21 @@
...
 
 # index dir will always have yml files only; but just in case
-for f in files:
-    if f.endswith(".yml"):
-        continue
-    files.remove(f)
+# for f in files:
+#    if f.endswith(".yml"):
+#        continue
+#    files.remove(f)
```
navidshaikh: Just remove the code instead of commenting out?

Issue #508, opened 2018-03-06
Registry UI title shows internal FQDN


bamachrn:
Provisioning with ansible scripts in prod results in showing internal FQDN.
We provision the registry with its internal ip address 
https://github.com/CentOS/container-pipeline-service/blob/master/provisions/roles/registry/tasks/main.yml#L26

This `public_registry` is the internal FQDN.  As a result of that it shows the internal FQDN after provision.

As a solution we run the UI pointing it to `registry.centos.org` and it works.
for a permanent fix to this we should add one more variable in the inventory file and point the UI to it.

PR #520, opened 2018-03-14
Fixed all the registry endpoints to get proper value in GUI

bamachrn:
Fixes: https://github.com/CentOS/container-pipeline-service/issues/508
Fixes: https://github.com/CentOS/container-pipeline-service/issues/574
This takes into consideration that
1. `public registry` is reachable by the internal builds. so in inventory file for deployment **other than** prod, it sholud be  `{jenkins_slave}`. And in prod it should be `registry.cento.org`
1. `intranet registry` is for insecure registry we push to. so in inventory fle has value `{jenkins_slave}:5000`

### DIFF
--- a/ci/lib.py
+++ b/ci/lib.py
@@ -273,7 +273,7 @@ scanner_worker
 
 [all:vars]
 db_host= {jenkins_master_host}
-public_registry= {jenkins_slave_host}
+public_registry= {jenkins_slave_host}:5000
 intranet_registry = {jenkins_slave_host}:5000
 copy_ssl_certs=True
 openshift_startup_delay=150
--- a/ci/lib.py
+++ b/ci/lib.py
@@ -273,7 +273,7 @@ scanner_worker
 
 [all:vars]
 db_host= {jenkins_master_host}
-public_registry= {jenkins_slave_host}:5000
+public_registry= {jenkins_slave_host}
 intranet_registry = {jenkins_slave_host}:5000
 copy_ssl_certs=True
 openshift_startup_delay=150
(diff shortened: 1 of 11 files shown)


bamachrn:
why we are seeing generic build failed. triggering the CI again.

bamachrn:
#dotests

bamachrn:
#dotests

bamachrn:
#dotests-debug

bamachrn:
@dharmit I checked it in devcloud it is fixing the logs issue #574. I rebased the code on master to make sure all the scanners run in devcloud.

provisions/roles/openshift/defaults/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 origin_image_registry: docker.io
 origin_image_name: openshift/origin
-origin_image_tag: v3.6.0
+origin_image_tag: v3.7.1
```
navidshaikh: this is new openshift origin version, can we rename the PR as "Upgrade using OpenShift version 3.7.1" ?

ci/lib.py
```diff
@@ -273,7 +273,7 @@ def generate_ansible_inventory(jenkins_master_host, jenkins_slave_host,
 
 [all:vars]
 db_host= {jenkins_master_host}
-public_registry= {jenkins_slave_host}
+public_registry= {jenkins_slave_host}:5000
```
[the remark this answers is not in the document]
bamachrn: yes, did few testing around this, registry is working fine, but ui is not coming up. fixing it.

provisions/roles/openshift/defaults/main.yml
bamachrn: that solved our current problem of not showing build logs, lets not rename it openshift upgrade, as it does not impact us much

bamachrn:
#dotests-debug

bamachrn:
#dotests-debug

PR #524, opened 2018-03-15
Removing gitpath and cccp.yml checks for prebuild.

mohammedzee1000:
#This removes checks for existence of pre-build and cccp.yml checks, if the prebuild parameters are specified.

### DIFF
--- a/testindex/validators.py
+++ b/testindex/validators.py
@@ -297,8 +297,26 @@ class IndexProjectsValidator(IndexValidator):
             # Else clone was success, check the git path
             git_path = clone_path + "/" + str(entry["git-path"])
 
+            # * Check for pre-build script
+            # TODO : Make a better implementation of pre-build script checking
+            # TODO : Ideally, if prebuild is not in entry, it wont reach here and this should happen if it is not None
+            prebuild_exists = False
+            if constants.PREBUILD_SCRIPT in entry and constants.PREBUILD_CONTEXT in entry:
+                prebuild_exists = True
+                prebuild_script = entry.get(constants.PREBUILD_SCRIPT)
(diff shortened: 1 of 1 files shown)


testindex/validators.py
```diff
@@ -297,8 +297,26 @@ def run(self):
...
 
+            # * Check for pre-build script
+            # TODO : Make a better implementation of pre-build script checking
+            # TODO : Ideally, if prebuild is not in entry, it wont reach here and this should happen if it is not None
+            prebuild_exists = False
+            if constants.PREBUILD_SCRIPT in entry and constants.PREBUILD_CONTEXT in entry:
+                prebuild_exists = True
+                prebuild_script = entry.get(constants.PREBUILD_SCRIPT)
+                prebuild_context = entry.get(constants.PREBUILD_CONTEXT)
+                if (prebuild_script and not path.exists(path.join(git_path, prebuild_script))
```
dharmit: That's a lot of `and` and `not` in there. Can we have a short comment about what exactly we're doing here?

dharmit:
@mohammedzee1000 

> This removes checks for existence of pre-build and cccp.yml, if the prebuild parameters are specified.

s/pre-build/git-path ?

bamachrn:
LGTM, merging

testindex/validators.py
navidshaikh: @mohammedzee1000 : Why didn't this being addressed? Shouldn't a complex operation be decorated with comment(s) ?

navidshaikh: @bamachrn : I would like to point out here. If we see a reviewer has asked for a change, we must ensure it is addressed. You merging it without review comments addressed, makes no point of a multiple reviewers and their reviews.

PR #527, opened 2018-03-19
Updating to skip build context check in case of pre-build.

mohammedzee1000:
One last check that needed to be disabled for prebuild containers, the check for build context.


bamachrn:
@navidshaikh would you please approve this one? we need to get this going to get the container-index PRs in.

PR #533, opened 2018-03-21
Remove rsync_ssh_opts from group_vars/all

cdrage:
If specified within group_vars/all, rsync_ssh_opts will override
whatever value is set in the "hosts" file.


navidshaikh:
#534 should just be enough then?

PR #534, opened 2018-03-21
Remove group_vars/all

cdrage:
It is now common practice to put all variables in the hosts /
inventory file rather than in group_vars/ (so you only have to edit one
file such as inventory or hosts.sample)

For example, some very-large Ansible projects using this characteristic:

https://github.com/kubernetes-incubator/kubespray
https://github.com/openshift/openshift-ansible

This PR removes group_vars/all and re-organized the hosts.sample file.


cdrage:
@navidshaikh Any idea why the tests are failing?

cdrage:
@navidshaikh Nvm, found the issue, will be pushing fix to: https://github.com/CentOS/container-pipeline-service/blob/b15ae4012d7ff1a05aed29a8de29602180cc962d/ci/lib.py#L293

cdrage:
#dotests

cdrage:
#dotests

PR #535, opened 2018-03-21
Remove Vagrantfile

cdrage:
Vagrant is no longer used / needed. Removes Vagrantfile and other
instances of Vagrant (hosts.vagrant and hosts.vagrant.minimal)


navidshaikh:
I'd suggest, we remove all the references of vagrant at once throughout the codebase. There are more if you check in `ci/cccp_ci.py`.

cdrage:
@navidshaikh I removed the files from `provisions` but not from the `server` folder as it looks like this is still being used by the CI (unless I'm mistaken?)

navidshaikh:
> it looks like this is still being used by the CI (unless I'm mistaken?)

We are not using Vagrantfiles anymore, not in CI as well.

PR #555, opened 2018-04-05
Uniforms scanners output filename to a pattern

navidshaikh:
Each atomic scanner will now export result file with name `<scanner_name>_results.json`, earlier scanners 
were exporting result filenames in different ways.


navidshaikh:
#dotests

PR #562, opened 2018-04-10
Fixes exporting result file in correct output dir for scanner-rpm-verify

navidshaikh:
There was a bug introduced while renaming the result filename, this changeset
  fixes it.


bamachrn:
LGTM, lets wait for CI to go green

PR #571, opened 2018-04-13
Uses devicemapper storage driver for devcloud environments

### LINKED ISSUE #567: Use devicemapper storage option as overlay2 doesn't support writable mounts (needed by pipeline-scanner)
Current/broken atomic RPM version:   `atomic-1.22.1-1.gitd36c015.el7.centos.x86_64`
Working atomic RPM version:              `atomic-1.20.1-9.git436cf5d.el7.centos.x86_64`

Broken functionality:
```
atomic mount -o rw $IMAGE $MOUNT_POINT
```

Issue:
- Earlier atomic mount command as mentioned above will mount the image under test in read write mode allowing pipeline scanners to create lock files in the mounted filesystem when it calls `yum` command.

- Now, after mounting the image under test, and running pipeline scanner, the filesystem is mounted in read only mode, not allowing `yum` to create lock file.

navidshaikh:
Fixes #567: We were using default storage driver overlay2 on devcloud
  environment, causing non writable mounts for pipeline-scanner.
  This changeset updates provisioning scripts to use devicemapper as
  storage driver when deployment environment is not in [production, pre-prod, staging].

### DIFF
--- a/provisions/roles/scanner/tasks/main.yml
+++ b/provisions/roles/scanner/tasks/main.yml
@@ -33,6 +33,15 @@
       state: permissive
   tags: scanner
 
+- name: Use devicemapper storage option for docker
+  replace: >
+    dest=/etc/sysconfig/docker-storage
+    regexp="^#*\s*DOCKER_STORAGE_OPTIONS=.*"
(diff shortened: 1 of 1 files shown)


dharmit:
Received emails on the alias. Looks good. Merging.

PR #576, opened 2018-04-20
[WIP] Uses registry as source of input for weekly-scan, not container-index

### LINKED ISSUE #559: Use registry catalog to find image list for weekly scan
Earlier, we used to go through the container-index and registry catalog to verify which images on the container-index exist on the registry catalog and then push them for weekly scan.

Now, we're just going through the container-index: https://github.com/CentOS/container-pipeline-service/blob/b23d106599ecab861ef7a85d5dadeef1d93fdf0b/jenkinsbuilder/weekly-scan.py#L56-L71

As a result of our current approach, we're trying to scan images that don't exist on the registry.

Either we should go back to how we were doing things earlier or just check the registry catalog when trying to perform a weekly scan.

navidshaikh:
Fix #559 Since listed projects in container-index might not be present in registry given failing builds etc.
Earlier container-index was source of input for weekly-scan.

### DIFF
--- a/container_pipeline/lib/inspect_registry.py
+++ b/container_pipeline/lib/inspect_registry.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python
+
+"""
+This module contains a class aimed for inspecting given docker
+registry using its REST APIs.
+It can
+ - list all images in registry
+ - filter images and list provided a namespace
+ - find tags of the repository
+"""
+
+import json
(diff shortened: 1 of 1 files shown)


navidshaikh:
Closing the issue as this PR was targeting the old architecture, and we have a new OpenShift based arch in place, where this is implemented from scratch.

PR #586, opened 2018-06-29
Adding remaining baseline validations to index ci

mohammedzee1000:
Note : This still skips on dependency validation, which i want to keep aside for now.

Issue #583, opened 2018-06-11
Reworking the index CI to make it more modular.


mohammedzee1000:
We need to re-write the index ci to make it more modular and easy to use.
Doing so will allow to 

1. Call validations individually from anywhere in the code.
2. Provide a list of validators to the index ci engine, which allows for plugging and unplugging of validations.
3. Reduction of index ci code base by atleast half.
4. Makes it easier to write test cases for index ci.

To achieve this, the following tasks need to be completed.

- [ ] Writing the basic framework of the index ci, including core component classes and helper functions.
- [ ] Writing checks for schema validation of index ci.
- [ ] Writing checks for value validation of the index ci.
- [ ] Writing CI for the new index CI
- [ ] Remove old Index CI and rewire the index CI.
- [ ] Discuss possibility of wiring index ci into the pipeline itself.

PR #588, opened 2018-07-03
Provisions having separate instance of state object

mohammedzee1000:
State was moved to callee level, instead of global. This allows for much more independent runs of index ci code and also easier CI for index ci implementation


mohammedzee1000:
Merge after https://github.com/CentOS/container-pipeline-service/pull/586

mohammedzee1000:

PR #590, opened 2018-07-04
Fixes validating projects against db from index in weekly scan

navidshaikh:
We must validate project entry in the db against every entry in the index before putting the weekly scan job on the tube.

This patchset 
 - skips the project from being put on the tube if the project does not exist in the db.
 - handles the DoesNotExist exception raised by Django, if the entry in the index does not match in projects table of the db.

### DIFF
--- a/container_pipeline/workers/scan.py
+++ b/container_pipeline/workers/scan.py
@@ -53,12 +53,12 @@ class ScanWorker(BaseWorker):
             self.logger.warning("Job data %s", str(self.job))
 
             self.set_buildphase_data(
-                build_phase_status='complete',
+                build_phase_status='failed',
                 build_phase_end_time=timezone.now()
             )
         else:
             self.set_buildphase_data(
-                build_phase_status='failed',
+                build_phase_status='complete',
                 build_phase_end_time=timezone.now()
             )
             self.logger.debug(str(scanners_data))
@@ -71,15 +71,17 @@ class ScanWorker(BaseWorker):
 
         # if weekly scan, push the job for notification
         if self.job.get("weekly"):
-            scanners_data["action"] = "notify_user"
-            self.queue.put(json.dumps(scanners_data), 'master_tube')
-            self.init_next_phase_data('delivery')
-            self.logger.debug(
-                str.format(
-                    "Weekly scan for {project} is complete.",
-                    project=self.job.get("namespace")
+            # send email of weekly scan only if scanners execution status=true
+            if status:
+                scanners_data["action"] = "notify_user"
+                self.queue.put(json.dumps(scanners_data), 'master_tube')
+                self.init_next_phase_data('delivery')
+                self.logger.debug(
+                    str.format(
+                        "Weekly scan for {project} is complete.",
+                        project=self.job.get("namespace")
+                    )
                 )
-            )
         else:
             # now scanning is complete, relay job for delivery
             # all other details about job stays same
--- a/jenkinsbuilder/weekly-scan.py
+++ b/jenkinsbuilder/weekly-scan.py
@@ -67,7 +67,14 @@ for f in files:
         desired_tag = entry["desired-tag"]
         email = entry["notify-email"]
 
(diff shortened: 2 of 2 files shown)


navidshaikh:
#dotests

jenkinsbuilder/weekly-scan.py
```diff
@@ -67,7 +67,14 @@
         desired_tag = entry["desired-tag"]
         email = entry["notify-email"]
 
-        entry_short_name = str(app_id) + "/" + str(job_id)
+        project_name = "{}-{}-{}".format(app_id, job_id, desired_tag)
+
+        try:
+            project = Project.objects.get(name=project_name)
+        except Project.DoesNotExist:
```
mohammedzee1000: :+1:  For specific exception handling but there should ideally be a top level exception block as well that can grab anything not handled by specific exception and handle it

navidshaikh: @mohammedzee1000 : in this case, the possible exception would be the `project_name` not found in the table, and it was the error which was blocking weeklyscan.

jenkinsbuilder/weekly-scan.py
```diff
@@ -109,18 +109,7 @@
...
 
         job = bs.put(json.dumps(data))
-        # Initializing Database entries
-        project = Project.objects.get(
-            name=project_name
-        )
-        if not project:
-            print(
-                str.format(
-                    "Skipping {}, as project object was not found",
```
navidshaikh: @mohammedzee1000 : also we were skipping the project before putting the job for same on tube!!  Its fixed now.

container_pipeline/workers/scan.py
```diff
@@ -53,12 +53,12 @@ def handle_job(self, job):
...
             self.set_buildphase_data(
-                build_phase_status='complete',
+                build_phase_status='failed',
                 build_phase_end_time=timezone.now()
             )
         else:
             self.set_buildphase_data(
-                build_phase_status='failed',
+                build_phase_status='complete',
                 build_phase_end_time=timezone.now()
```
bamachrn: does this mean we were having it wrong? thanks for fixing this up.

container_pipeline/workers/scan.py
```diff
@@ -71,15 +71,17 @@ def handle_job(self, job):
...
+            if status:
+                scanners_data["action"] = "notify_user"
+                self.queue.put(json.dumps(scanners_data), 'master_tube')
+                self.init_next_phase_data('delivery')
+                self.logger.debug(
+                    str.format(
+                        "Weekly scan for {project} is complete.",
+                        project=self.job.get("namespace")
+                    )
                 )
```
bamachrn: so we are not sending any notification if it weekly scan and scanning failed, right?

jenkinsbuilder/weekly-scan.py
bamachrn: :+1:

container_pipeline/workers/scan.py
navidshaikh: Right.

For normal build scan failures:

Since the `scan` stage in normal build scenario should not affect final delivery of the image, we move on and deliver the image, that is next [else](https://github.com/CentOS/container-pipeline-service/pull/590/files#diff-92601f9830cf93227125d7e969bc425aR85) statement. The scanners execution in most of th cases will fail only if the image is not present in registry, if the image is built, it will be scan.
In case, where the image is built, but scan stage is failed, the image will be delivered and the email will have
image build details and no scan details.

container_pipeline/workers/scan.py
navidshaikh: hah! yes, we were having it wrong earlier, fixed now. :+1:

navidshaikh:
> assuming we are not getting into the loop of empty content for normal scans

@bamachrn  No, the loop of emails is fixed already, we also fixed empty contents with this PR.

navidshaikh:
#dotests

bamachrn:
LGTM, will merge once CI is green

PR #591, opened 2018-07-04
Adding more tests for the index CI

mohammedzee1000:
This PR is part 2 of adding tests for index CI. As the number of tests increase, they have been broken down.

Merge after : https://github.com/CentOS/container-pipeline-service/pull/587

PR #593, opened 2018-07-09
Adding init for new index ci and adding tests for the same

mohammedzee1000:
This PR achieves 

1. Adding init with argument parsing and summary printing for new index ci
1. Adding tests for init.

### DIFF
--- a/ci/container_index/engine.py
+++ b/ci/container_index/engine.py
@@ -32,12 +32,18 @@ class Engine(object):
     def __init__(
             self, schema_validators=None,
             value_validators=None, index_location="./", verbose=True,
-            the_state=None
+            skip_schema=False, skip_value=False, the_state=None
     ):
         """
         Initializes the test engine
         """
         self.verbose = verbose
+        self.skip_schema = skip_schema
+        self.skip_value = skip_value
+
+        if self.skip_schema and self.skip_value:
+            raise Exception("Cannot skip through all tests")
+
         # Index file location needs to be provided
         if not path.exists(index_location):
             raise Exception(
@@ -60,24 +66,26 @@ class Engine(object):
         # Collect the validators that need to run.
         self.validators = []
         # - Schema Validators:
-        if schema_validators != "None":
+        if schema_validators:
             if (not (schema_validators and
                      isinstance(schema_validators, list)) or
                     len(schema_validators) <= 0):
                 v_list = config.schema_validators
             else:
                 v_list = schema_validators
-            self._load_validators(schema_validation, v_list)
+            if not self.skip_schema:
+                self._load_validators(schema_validation, v_list)
 
         # - Value Validators
-        if value_validators != "None":
+        if value_validators:
             if (not value_validators or not
                     isinstance(value_validators, list) or
                     len(value_validators) <= 0):
                 v_list = config.value_validators
             else:
                 v_list = value_validators
-            self._load_validators(value_validation, v_list)
+            if not self.skip_value:
+                self._load_validators(value_validation, v_list)
 
         self.summary = {}
         self.state = the_state if the_state else state.State()
--- a/ci/container_index/engine.py
+++ b/ci/container_index/engine.py
@@ -60,7 +60,7 @@ class Engine(object):
         # Collect the validators that need to run.
         self.validators = []
         # - Schema Validators:
-        if schema_validators:
+        if schema_validators != "None":
             if (not (schema_validators and
                      isinstance(schema_validators, list)) or
                     len(schema_validators) <= 0):
@@ -70,7 +70,7 @@ class Engine(object):
             self._load_validators(schema_validation, v_list)
 
         # - Value Validators
-        if value_validators:
+        if value_validators != "None":
             if (not value_validators or not
                     isinstance(value_validators, list) or
                     len(value_validators) <= 0):
--- a/ci/container_index/run.py
+++ b/ci/container_index/run.py
@@ -1,0 +1,91 @@
+import sys
+from os import path
+
+fl = path.dirname(path.realpath(__file__))
+sys.path.append(path.join(fl, "..", ".."))
+
+from argparse import ArgumentParser
+import ci.container_index.engine as engine
+
+
+def init_parser():
+    parser = ArgumentParser(
+        description="This tool lets you run index validations from the CLI"
+    )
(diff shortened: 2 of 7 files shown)


mohammedzee1000:
Merge after https://github.com/CentOS/container-pipeline-service/pull/591

mohammedzee1000:

ci/container_index/run.py
```diff
@@ -0,0 +1,115 @@
+import sys
+from os import path
+
+fl = path.dirname(path.realpath(__file__))
+sys.path.append(path.join(fl, "..", ".."))
```
navidshaikh: Add some comments here about what module needs to be added to the path.
Can we avoid this modification of sys.path in run.py and have the module path added at the starter script?  Like you can specify `PYTHONPATH` while invoking the CI initi.

ci/container_index/run.py
```diff
@@ -0,0 +1,115 @@
...
+        index_location=index,
+        verbose=verbose,
+        schema_validators=schema_validators,
+        value_validators=value_validators
+    ).run()
+    print_summary(summary, verbose)
+    if not success:
+        exit(1)
+
+
```
navidshaikh: add `if __name__ == __main__`: here may be?

ci/container_index/engine.py
```diff
@@ -59,23 +60,27 @@ def __init__(
         # Collect the validators that need to run.
         self.validators = []
         # - Schema Validators:
-        if (not (schema_validators and isinstance(schema_validators, list)) or
-                len(schema_validators) <= 0):
-            v_list = config.schema_validators
-        else:
-            v_list = schema_validators
-        self._load_validators(schema_validation, v_list)
+        if schema_validators != "None":
```
navidshaikh: What would be the scenario for `schema_validators="None"` ?

navidshaikh: as in, its value would be string `"None"` and not `None` ?

ci/container_index/run.py
```diff
@@ -0,0 +1,113 @@
...
+def init_parser():
+    parser = ArgumentParser(
+        description="This tool lets you run index validations from the CLI"
+    )
+    parser.add_argument(
+        "-i",
+        "--index",
+        help="Specify location of index. This is where index.d is Default .",
```
navidshaikh: *`is default.`

navidshaikh: *`Specify location of container index.`

ci/container_index/run.py
```diff
@@ -0,0 +1,113 @@
...
+        metavar="INDEX_PATH",
+        default="."
+    )
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        help="Verbose mode prints more output of the tests",
```
navidshaikh: *`tests.`

ci/container_index/run.py
```diff
@@ -0,0 +1,113 @@
...
+        "--schemavalidators",
+        help="Comma seperated list of schema validators to run. Use 'None'"
+             " to skip all",
+        metavar="SCHEMA_VALIDATORS"
+    )
+    parser.add_argument(
+        "-u",
+        "--valuevalidators",
+        help="Comma seperated list of value validators to run. Use 'None' to "
+             "skip all",
```
navidshaikh: Should the choice be to not specify it for assuming `None`? (it relates to above asked question for value `"None"`)

ci/container_index/run.py
```diff
@@ -0,0 +1,113 @@
...
+    index = args.index
+    verbose = True if args.verbose else False
+    schema_validators = None if not args.schemavalidators else (
+        'None' if 'None' in args.schemavalidators else
+        args.schemavalidators.split(',')
+    )
+    value_validators = None if not args.valuevalidators else (
+        'None' if 'None' in args.valuevalidators else
+        args.valuevalidators.split(',')
+    )
```
navidshaikh: I think we can simplify this by eliminating `"None"` input for *validators.

ci/container_index/engine.py
mohammedzee1000: If you don't want any of the schema validations to happen for example. Maybe i should make it as some other value like skip? and use None to say use default?

navidshaikh: If there is no option specified at CLI, it should use the default. Why would a user want to `skip` it?

ci/container_index/run.py
mohammedzee1000: Sorted

ci/container_index/run.py
mohammedzee1000: Sorted

ci/container_index/run.py
mohammedzee1000: Fixed

ci/container_index/run.py
```diff
@@ -0,0 +1,125 @@
...
+        "-v",
+        "--verbose",
+        help="Verbose mode prints more output of the tests.",
+        action="store_true"
+    )
+    parser.add_argument(
+        "-s",
+        "--schemavalidators",
+        help="Comma seperated list of schema validators to run."
+             " to skip all",
```
navidshaikh: I think the help text needs an update?

navidshaikh: I think the help text needs an update?

` to skip all` ?

ci/tests/test_01_functional/test_04_index_ci/__init__.py
```diff
@@ -134,7 +135,119 @@ def test_04_indexci_runs_specific_value_validators(self):
...
+                }
+            }
+        )
+        cmd = [
+            "python",
+            str(path.join(
+                path.dirname(path.realpath(__file__)),
+                "..",
+                "..",
+                "..",
```
navidshaikh: Can you think of a way to avoid this ?

ci/container_index/run.py
mohammedzee1000: Sorted

ci/tests/test_01_functional/test_04_index_ci/__init__.py
mohammedzee1000: I will definitely try to figure out a way to avoid this.
But for now, as interim, we should be ok?

mohammedzee1000:
Strange, did the pythonpath thing not get applied yet to ci job?

PR #595, opened 2018-07-17
Setting pythonpath before running the CI.

mohammedzee1000:
The pythonpath needs to be set to root of service code
or imports will fail.


mohammedzee1000:

PR #596, opened 2018-07-25
Adding more explicit setting of pythonpath for index ci.

mohammedzee1000:
PYTHONPATH is becoming an issue, hence, setting it more explicitly and adding printing of the path.


ci/cccp_ci_container_index.sh
```diff
@@ -7,7 +7,7 @@ ssh_cmd="ssh $sshopts $CICO_hostname"
 $ssh_cmd "yum -y install epel-release && yum -y install rsync git PyYAML python-networkx"
 rsync -e "ssh $sshopts" -Ha $(pwd)/ $CICO_hostname:payload
 $ssh_cmd git clone https://github.com/CentOS/container-pipeline-service.git 
-$ssh_cmd PYTHONPATH="`pwd`/container-pipeline-service" /usr/bin/python container-pipeline-service/ci/container_index/run.py  -vi ./payload
+$ssh_cmd export PYTHONPATH="`pwd`/container-pipeline-service" && echo $PYTHONPATH && /usr/bin/python container-pipeline-service/ci/container_index/run.py  -vi ./payload
```
navidshaikh: I think adding before command without export is just fine too.

mohammedzee1000: That is what i tried, and it seems to be failing so im exporting and echoing it to be sure.

PR #610, opened 2018-09-06
Converts pipeline name to lower case

navidshaikh:
Fix OSIO 1522

Since OpenShift doesn't allow the value of metadata.name to have uppercase chars.

Added a regular expression to validate the pipeline name generated
for using in template's metadata.name field. oc imposes certain guidelines
for same and provides the regular expression in error/help message to be
used for validation in case it found some invalid string.

We are using same regular expression as provided by oc to validate
the pipeline name while processing index and generating build configs
from it.

Added a custom exception, which will be raised if the populated string
is invalid.

Added a unit test as well to test the functionality.


navidshaikh:
created against wrong branch, closing it.

PR #612, opened 2018-09-06
Processes the index projects in batches

navidshaikh:
This changeset updates index reader to process index-projects
in defined (default=5) length of batches. Index reader, splits
the index projects in batches and processes each batch one by one.

In between processing batches of projects, it checks if there are
any running builds. If it finds a running build, it waits for given
poll_cyle (default=120 seconds) and checks again. As it finds, openshift
idle (not processing any builds ATM), it queues the batch and waits for
its completion as above.

The weekly scan projects are processed at the end after processing build
jobs completely. Every weekly scan job is processed with 5 seconds delay
in between.

Sample output first run of the index

```
[pipeline-service] Running shell script
+ python ccp/index_reader.py /tmp/container-index/index.d 172.29.33.53:5000 cccp nshaikh@redhat.com smtp://mail.centos.org
Number of projects in index 9
Number of projects in OpenShift 0
Number of projects to be updated/created: 9
Processing projects batch: ['test-anomaly-latest', 'test-python-release', 'dharmit-freegeoip-latest', 'dharmit-nsq-latest', 'mattermost-mattermost-github-integration-92394f2']
buildconfig.build.openshift.io "test-anomaly-latest" created

buildconfig.build.openshift.io "test-python-release" created

buildconfig.build.openshift.io "dharmit-freegeoip-latest" created

buildconfig.build.openshift.io "dharmit-nsq-latest" created

buildconfig.build.openshift.io "mattermost-mattermost-github-integration-92394f2" created

Processing projects batch: ['arrfab-nanoc-latest', 'openshift-origin-base-latest', 'openshift-origin-source-latest', 'centos-centos-latest']
buildconfig.build.openshift.io "arrfab-nanoc-latest" created

buildconfig.build.openshift.io "openshift-origin-base-latest" created

buildconfig.build.openshift.io "openshift-origin-source-latest" created

buildconfig.build.openshift.io "centos-centos-latest" created

Processing weekly scan projects..
buildconfig.build.openshift.io "wscan-test-anomaly-latest" created

buildconfig.build.openshift.io "wscan-test-python-release" created

buildconfig.build.openshift.io "wscan-dharmit-freegeoip-latest" created

buildconfig.build.openshift.io "wscan-dharmit-nsq-latest" created

buildconfig.build.openshift.io "wscan-mattermost-mattermost-github-integration-92394f2" created

buildconfig.build.openshift.io "wscan-arrfab-nanoc-latest" created

buildconfig.build.openshift.io "wscan-openshift-origin-base-latest" created

buildconfig.build.openshift.io "wscan-openshift-origin-source-latest" created

buildconfig.build.openshift.io "wscan-centos-centos-latest" created
```


navidshaikh:
This PR is rebased on #617, which needs to go first.

ccp/index_reader.py
```diff
@@ -188,7 +184,7 @@ def read_yaml(self, filepath):
             with open(filepath) as fin:
                 data = yaml.load(fin, Loader=yaml.BaseLoader)
         except yaml.YAMLError as exc:
-            print ("Failed to read {}".format(filepath))
+            _print("Failed to read {}".format(filepath))
             raise(exc)
         else:
```
bamachrn: how are we handling this `raise` here we are raising exception from exception?

ccp/index_reader.py
```diff
@@ -6,25 +6,21 @@
...
-class ErrorAccessingIndexEntryAttributes(Exception):
+def _print(msg):
     """
-    Exception to be raised when there are errors accessing
-    index entry attributes
+    _prints the given msg
     """
-    pass
+    print (msg)
+    sys.stdout.flush()
```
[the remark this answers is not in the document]
navidshaikh: That's correct.

ccp/index_reader.py
navidshaikh: We are raising the exception but ensuring we print error message along with filepath with issue.

navidshaikh: We have not handled this. Index Reader expects the yaml files to be in sane format. Do we want to log error and proceed further in this case as well?

ccp/index_reader.py
```diff
@@ -379,26 +374,66 @@ def start_build(self, pipeline_name):
...
+
+        command = """\
+oc get builds -o name -o template --template='{{range .items }} \
+{{if and %s }} {{.metadata.name}}:{{.status.phase}} \
+{{end}}{{end}}'""" % condition
+        output = run_cmd(command, shell=True)
+        output = output.strip().split(' ')
+        output = [each for each in output
+                  if not each.startswith(tuple(filter_builds))
+                  and each]
```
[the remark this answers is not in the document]
navidshaikh: Ack!

ccp/index_reader.py
bamachrn: yeah, these should be taken care of by the index CI. But index reader should not fail entirely for single malformed entry. I would suggest we log it and continue to process other entries.

navidshaikh: Agreed. I'll update the PR.

navidshaikh:
Rebased and fixed the review comments.

ccp/index_reader.py
```diff
@@ -449,33 +502,116 @@ def run(self):
...
+            # wait until current outstanding builds are more than
+            # configured cap for outstanding builds
+            while len(outstanding_builds) > batch_outstanding_builds_cap:
+                _print("Waiting for completion of builds {}\n".format(
+                    outstanding_builds))
+                time.sleep(polling_interval)
+
+                outstanding_builds = self.bc_manager.list_builds_except(
+                    status=["Complete", "Failed"],
+                    filter_builds=self.infra_projects)
```
dharmit: IIUC, we're executing a shell command to get output here. `run_cmd` and `list_builds_except` don't seem to have exception handling. So, if fetching outstanding builds fails due to some issue (like failure in executing `oc` command), seed-job would fail. Correct me if I'm wrong.

If I'm correct, I think this part is deep into the execution of our service. We could give a few retries to the scenario where `oc` fails to get our desired task done and fail the seed-job appropriate message if it fails even after the retries.

Thoughts @navidshaikh @bamachrn @mohammedzee1000 ?

navidshaikh: I think it makes sense - We can keep a retry cycle of 30 seconds with timeout limit set to 5 minutes ? And fail it is 5 minutes (10 retries) are lapsed ?

navidshaikh: I am adding the retries in the actual method itself, as its being called multiple times.

navidshaikh: It makes sense. It applies to all the operations performed via `oc`. Adding a retry decorator which can be re-used and have it applied to al oc related operations.

PR #614, opened 2018-09-06
Adds provision to provide exception jobs list while removing stale jobs

navidshaikh:
This is to cater requiremenet of functional CI feature, where
we have added a ci-job. This job is not part of container index
and thus as per principal found stale job when index reader runs.

While finding stale jobs, now we are adding provision to provide
method with exception-jobs list, which will be exempted while finding
stale jobs (which eventually gets removed from OpenShift).
As we don't want ci-job to be removed from OpenShift after index-reader
(or seed-job) is run. The default value for exception_jobs parameter
is ["ci-job"].

### DIFF
--- a/ccp/index_reader.py
+++ b/ccp/index_reader.py
@@ -449,18 +449,24 @@ class Index(object):
     def __init__(self, index, registry_url, namespace,
                  from_address, smtp_server,
                  ccp_openshift_slave_image,
-                 notify_cc_emails):
+                 notify_cc_emails,
+                 infra_projects=["seed-job"],
+                 ci_projects=["ci-success-job", "ci-failure-job"]):
+
         # create index reader object
         self.index_reader = IndexReader(index, namespace)
+
         # create bc_manager object
         self.bc_manager = BuildConfigManager(
             registry_url, namespace, from_address, smtp_server,
             ccp_openshift_slave_image, notify_cc_emails)
-        self.infra_projects = ["seed-job"]
 
-    def find_stale_jobs(self,
-                        oc_projects, index_projects,
-                        exception_jobs=["ci-job"]):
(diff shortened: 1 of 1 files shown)


ccp/index_reader.py
```diff
@@ -366,14 +367,29 @@ def __init__(self, index, registry_url, namespace,
             registry_url, namespace, from_address, smtp_server)
         self.infra_projects = ["seed-job"]
 
-    def find_stale_jobs(self, oc_projects, index_projects):
+    def find_stale_jobs(self,
+                        oc_projects, index_projects,
+                        exception_jobs=["ci-job"]):
         """
```
bamachrn: we need to add `ci-success-job` and `ci-failure-job` in the exception list.

ccp/index_reader.py
```diff
@@ -366,14 +367,29 @@ def __init__(self, index, registry_url, namespace,
...
 
-        return list(set(oc_projects) - set(index_projects))
+        stale_jobs = list(set(oc_projects) - set(index_projects))
+
+        if stale_jobs:
+            # remove the exception list from found stale jobs
+            print ("Removing exception job(s) {} from found stale jobs, "
+                   "mentioned jobs won't be removed OpenShift..".format(
+                       exception_jobs))
+            return list(set(stale_jobs) - set(exception_jobs))
```
bamachrn: :+1:

navidshaikh:
Rebased and updated as per review comments.

navidshaikh:
@bamachrn: The PR is updated as per your review, can you take a look at updates to PR?

navidshaikh:
#dotests-unittests

bamachrn:
LGTM.

PR #617, opened 2018-09-06
Continues processing index entries upon errors, updates buildconfig names to lower case

navidshaikh:
1. 
Updated the PR to change the behavior of seed-job as well.
Now the seed-job will continue to process rest of the entries in index if it encountered errors while processing index. It will print the error and move on.

2.
Fix OSIO 1522
Since OpenShift doesn't allow the value of metadata.name to have uppercase chars.
Added a regular expression to validate the pipeline name generated
for using in template's metadata.name field. oc imposes certain guidelines
for same and provides the regular expression in error/help message to be
used for validation in case it found some invalid string.

We are using same regular expression as provided by oc to validate
the pipeline name while processing index and generating build configs
from it.

Added a custom exception, which will be raised if the populated string
is invalid.

Added a unit test as well to test the functionality.

### DIFF
--- a/ccp/index_reader.py
+++ b/ccp/index_reader.py
@@ -3,11 +3,21 @@ This script parses the container index specified and
 creates the Jenkins pipeline projects from entries of index.
 """
 
+import exceptions
+import re
 import subprocess
 import sys
 import yaml
 
 from glob import glob
+
+
+class InvalidPipelineName(exceptions.Exception):
+    """
+    Exception to be raised when pipeline name populated doesn't
+    confornt to allowed value for openshift template field metadata.name
+    """
+    pass
 
 
 def run_cmd(cmd, shell=False):
@@ -129,8 +139,24 @@ class Project(object):
         Returns the pipeline name based on appid, jobid and desired_tag
         and also converts it to lower case
         """
-        return "{}-{}-{}".format(
+        pipeline_name = "{}-{}-{}".format(
             self.app_id, self.job_id, self.desired_tag).lower()
+
+        # pipeline name which becomes value for metadata.name field in template
+        # must confront to following regex as per oc
+        # We tried to make the string acceptable by converting it to lower case
+        # Below we are adding another gate to make sure the pipeline_name is as
+        # per requirement, otherwise raising an exception with proper message
+        # to indicate the issue
+        pipeline_name_regex = ("^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]"
+                               "([-a-z0-9]*[a-z0-9])?)*$")
+        match = re.match(pipeline_name_regex, pipeline_name)
+
+        if not match:
+            msg = ("The pipeline name populated {} can't be used in OpenShift "
+                   "template in metadata.name field. ".format(pipeline_name))
+            raise(InvalidPipelineName(msg))
+        return pipeline_name
 
 
 class IndexReader(object):
@@ -242,7 +268,7 @@ class BuildConfigManager(object):
         """
         Applies the build job template that creates pipeline to build
         image, and trigger first time build as well.
-        :param project: The name of project,where the template is to be applied
+        :param project: The name of project, where template is to be applied
         :param template_location: The location of the template file.
         """
         oc_process = "oc process -f {0} {1}".format(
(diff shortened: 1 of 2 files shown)


ccp/index_reader.py
```diff
@@ -3,13 +3,23 @@
...
+import exceptions
+import re
 import subprocess
 import sys
 import yaml
 
 from glob import glob
 
 
+class InvalidPipelineName(exceptions.Exception):
```
dharmit: Should we create a separate file for exceptions? We might add more exceptions in future now that we've become serious about testing. In that case we can do something like `from ccp_exceptions import InvalidPipelineName`. 

Another reason for this is that we will only keep increasing the size of py files if we add all into one. 

This is just a comment, I leave the final call on you @navidshaikh

ccp/index_reader.py
```diff
@@ -126,9 +136,27 @@ def load_project_entry(self, entry):
...
+        # per requirement, otherwise raising an exception with proper message
+        # to indicate the issue
+        pipeline_name_regex = ("^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]"
+                               "([-a-z0-9]*[a-z0-9])?)*$")
+        match = re.match(pipeline_name_regex, pipeline_name)
+
+        if not match:
+            msg = ("The pipeline name populated {} can't be used in OpenShift "
+                   "template in metadata.name field. ".format(pipeline_name))
+            raise(InvalidPipelineName(msg))
```
dharmit: From the definition of `InvalidPipelineName`, it seems to me that if it's going to encounter a project where this `raise` statement indeed gets executed, it will cause the seed-job to fail with an exception. Am I seeing it correctly? I think we would prefer having a traceback and continue execution of seed-job so that other builds in the batch don't get affected.

dharmit: Adding the `InvalidPipelineName` exception is a part of taking care of buildconfig name. So if that exception arises, shouldn't that be handled in the same PR? I believe it should be handled in this PR only.

bamachrn: where are we catching the exception? if we dont have a catch it can potentially restart jenkins. yeah I would also vote for adding the exception handling in the same PR.

navidshaikh:
Updated the PR to change the behavior of seed-job as well.
Now the seed-job will continue to process rest of the entries in index if it encountered errors while processing index. It will print the error and move on.

navidshaikh:
Added 3 new unit tests to test the functionality. Sample run
```
nosetests -w . -v tests 
IndexReader: Tests processing pipeline_name ... ok
IndexReader: Tests processing pipeline_name for upper case params ... ok
IndexReader: Tests exception while processing pipeline name ... ok
Tests creating Project class object which loads entries ... ok
IndexReader: Tests exception while loading invalid index entry ... ok
Test processing pre_build_context ... ok
Test processing pre_build_script ... ok
Tests depends_on field processing and defaults ... ok
Tests desired tag processing and defaults ... ok
Test the helper method to replace . / : _ with hyphen ... ok
Test __str__ representation of Project object ... ok
Notifications: Tests processing body of email ... ok
Notitications: Tests processing subject of email ... ok

----------------------------------------------------------------------
Ran 13 tests in 0.002s

OK
```

navidshaikh:
Tested the functionality by explicitly malforming an index entry (removed target-file [here](https://github.com/navidshaikh/ccp-openshift-index/commit/9885165ffdde465c2c17d0d399d4b027d1a8cd4b#diff-b1cfebf5d6f069b7b72ff4a753400ab3L8)). Sample output of seed-job run

```
+ python ccp/index_reader.py /tmp/container-index/index.d 172.29.33.53:5000 cccp nshaikh@redhat.com smtp://mail.centos.org
Error processing index entry {u'app-id': u'centos', u'git-url': u'https://github.com/sclorg/s2i-perl-container.git', u'git-path': u'5.20/', u'build-context': u'./', u'depends-on': [u'centos/centos:latest', u'centos/s2i-core-centos7:latest', u'centos/s2i-base-centos7:latest'], u'notify-email': u'shaikhnavid14@gmail.com', u'desired-tag': u'latest', u'git-branch': u'master', u'job-id': u'perl-520-centos7', u'id': u'1'}. Moving on.
Error: 'target-file'
Number of projects in index 9
Number of projects in OpenShift 0
Number of projects to be updated/created: 9
```

PR #621, opened 2018-09-07
Updates container-capabilities scanner to understand 'null' value

navidshaikh:
Fix OSIO #1535 : docker inspect will return `null` value if RUN label is absent in image under test.
Container capabilities scanner was taking `null` value as RUN label value. Updated the scanner to understand null value and categorize it as ` RUN label is not available in image under test.`

Also, updated the `CMD` instruction of scan-data container. Now it's `/bin/sleep infinity`.

### DIFF
--- a/Dockerfiles/ccp-openshift-scan/Dockerfile
+++ b/Dockerfiles/ccp-openshift-scan/Dockerfile
@@ -4,4 +4,4 @@ ADD scanning /opt/scanning
 
 VOLUME ["/opt/scanning"]
 
-CMD ["/bin/tail", "-f", "/dev/null"]
+CMD ["/bin/sleep", "infinity"]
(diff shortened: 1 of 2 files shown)


Dockerfiles/ccp-openshift-scan/Dockerfile
```diff
@@ -4,4 +4,4 @@ ADD scanning /opt/scanning
 
 VOLUME ["/opt/scanning"]
 
-CMD ["/bin/tail", "-f", "/dev/null"]
+CMD ["/bin/sleep", "infinity"]
```
dharmit: What's the problem with `tail -f /dev/null`? Is `sleep infinity` a better approach? Is there a reference for it?

dharmit: I tried running both the commands and used `htop` to observe them. `tail` is not showing the behavior of continuous processing. I'm saying so because CPU column doesn't show any spike. I even did

```bash
$ for i in `seq 100`; do echo $i > /dev/null; sleep 1; done
``` 
to see if there's a spike. But there wasn't any.

It's how we think tail works that's probably making us believe so.

PR #626, opened 2018-09-11
Fixes bugs in scanners

navidshaikh:
This changeset fixes bugs in scanners.

 1. Uses yum python API client library to list yum updates
    
      Stops using system commands and parsing stdout, now uses
      yum python API client library to list and print the updates
      in cleaner way.
    
     Uses a Suppressor for not printing the yum repo updates stdout
     before printing results.
    
     Adds ability to list installed version of package, for which
     updates are available.
     Adds ability to list the source yum repo.

 2. Adds error checking in rpmverify scanner
    
    Adds missing sys import
    Adds formating indexes to be compatible for python version in
    centos6 images.

 3. Fixes string formatting in misc and capabilities scanner
    
      Fixes string formatting
      Uses explicit indexes in string formatting for python version
      in centos6 image

4.  Uses /usr/bin/python as entrypoint to stay compatible with centos6 images
    
      Since python is installed at /usr/bin/python in CentOS6 image,
      and earlier we were using /bin/python , which is unavailable in
      CentOS6 image.
      CentOS7 image has needed symlinks to stay compatible with either
      /usr/bin/python or /usr/bin/python.

### DIFF
--- a/Dockerfiles/ccp-openshift-scan/scanning/container-capabilities.py
+++ b/Dockerfiles/ccp-openshift-scan/scanning/container-capabilities.py
@@ -46,7 +46,7 @@ def check_args(cmd):
             if not found_sec_arg:
                 print("\nThis container uses privileged "
                       "security switches:")
-            print("\nINFO: {}\n\t{}".format(sec_arg, security_args[sec_arg]))
+            print("\nINFO: {0}\n\t{1}".format(sec_arg, security_args[sec_arg]))
             found_sec_arg = True
     if found_sec_arg:
         print("\nFor more information on these switches and their "
@@ -69,7 +69,7 @@ if __name__ == "__main__":
     if len(sys.argv) != 2:
         example = ('python container-capabilities.py'
                    ' "docker run --privileged $IMAGE /bin/true"')
-        print ("Please provide one argument as\n{}".format(example))
+        print ("Please provide one argument as\n{0}".format(example))
         sys.exit(1)
 
     try:
@@ -77,5 +77,5 @@ if __name__ == "__main__":
         run_scan(cli_arg)
     except Exception as e:
         print ("Error occurred in Container Capabilities scanner execution.")
-        print ("Error: %s".format(e))
+        print ("Error: {0}".format(e))
         sys.exit(1)
--- a/Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
+++ b/Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
@@ -1,44 +1,117 @@
-# scan script for yum list update scannr
+# scan script for yum list update scanner
 
-import scan_lib
+import sys
+import yum
 
 
-def yum_updates():
+class SysStdoutSuppressor(object):
     """
-    Finds yum updates
+    Class which can be used to mute output printed
+    on sys.stdout.
+    It just mutes the sys.stdout and takes care of raising
+    the exception if encountered in calling method.
     """
-    command = ["yum", "-q", "check-update"]
-    out, err = scan_lib.run_cmd_out_err(command)
-    return out, err
+
+    def __enter__(self):
+        self.stdout = sys.stdout
+        sys.stdout = self
+
+    def __exit__(self, type, value, traceback):
+        # revert back sys.stdout to original form before exiting
+        sys.stdout = self.stdout
+        if type is not None:
+            raise(value)
+
+    def write(self, x):
+        """
+        Overriding sys.stdout.write, this method won't print
+        on stdout
+        """
+        pass
 
 
-def find_updates():
+class YumUpdates(object):
     """
-    process yum updates
+    Class with related methods to find the yum updates
+    available using python yum API client
     """
-    out, err = yum_updates()
-    if err:
-        print ("Error:")
-        print ("Error retrieving yum updates.\n{}".format(err))
-        return
-    out = out.strip()
 
-    if not out:
-        print ("No yum updates required.")
-        return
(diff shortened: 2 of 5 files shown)


Dockerfiles/ccp-openshift-scan/scanning/container-capabilities.py
```diff
@@ -46,7 +46,7 @@ def check_args(cmd):
             if not found_sec_arg:
                 print("\nThis container uses privileged "
                       "security switches:")
-            print("\nINFO: {}\n\t{}".format(sec_arg, security_args[sec_arg]))
+            print("\nINFO: {0}\n\t{1}".format(sec_arg, security_args[sec_arg]))
```
navidshaikh: The python version available in CentOS6 image does not understanding `{} {}`, one needs to provide explicit ordering.

When I fixed a scanner, I tested the fixes and all other scanners with CentOS6 and CentOS7 images to be sure it works as expected, I found this issue. We didn't hit this issue in testing because the execution failed before code could reach until here, the execution failed at the very start for python's absolute path.

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
```diff
@@ -1,44 +1,117 @@
-# scan script for yum list update scannr
+# scan script for yum list update scanner
 
-import scan_lib
+import sys
+import yum
 
 
-def yum_updates():
+class SysStdoutSuppressor(object):
```
navidshaikh: When we were doing yum updates checking we're providing `-q` option to be quiet for yum repo updates download stdout. This is same implementation in python. We don't want to print anything other than the updates available, helps the logs fetching to do less processing.

navidshaikh: If we don't do this, the python processing for finding updates will print on stdout.

navidshaikh: Basically, without this, its `yum check-update`

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
```diff
@@ -1,44 +1,117 @@
...
-    try:
-        find_updates()
-    except Exception as e:
-        print ("Error occurred in RPM Updates scanner execution.")
-        print ("Error: %s".format(e))
-        sys.exit(1)
+    y = YumUpdates()
+    # prevent repo update downloads stdout
+    with SysStdoutSuppressor():
+        updates = y.find_updates()
```
navidshaikh: The suppressor class is used here, we dont want to print extra output while fetching available updates.

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
navidshaikh: https://www.python.org/dev/peps/pep-0343/ this should help review this.

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
```diff
@@ -1,44 +1,117 @@
...
-    if not out:
-        print ("No yum updates required.")
-        return
+    def __init__(self):
+        """
+        Instantiate
+        :arg needed_fields: Properties of RPM object  while processing
+        """
+        self.yum_obj = yum.YumBase()
+        self.needed_fields = ["name", "vra", "repo"]
```
navidshaikh: this variable is not needed

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
```diff
@@ -1,44 +1,117 @@
...
+
+    def find_updates(self):
+        """
+        Find yum updates and returns the needed_fields attributes of RPM
+        as dictionary for available RPM updates. Returns empty dictionary
+        if no updates are required.
+
+        :return: List of dictionaries of needed_fields as keys and value
```
navidshaikh: this docstring needed an update

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
```diff
@@ -1,44 +1,115 @@
...
+
+    def print_updates(self, updates):
+        """
+        Print the updates on stdout
+        :arg updates: Updates provided as list of dictionaries,
+                      the output received from find_updates() method
+        :type updates: List
+        """
+        if not updates:
+            print("No RPM updates available as per configured yum repos.")
```
dharmit: This sounds more appropriate to me:

> No RPM updates available in the configured yum repos.

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
```diff
@@ -1,44 +1,117 @@
...
+        on stdout
+        """
+        pass
+
+
+class YumUpdates(object):
     """
-    process yum updates
+    Class with related methods to find the yum updates
+    available using python yum API client
```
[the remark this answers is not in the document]
navidshaikh: ok -
I am updating it to
```
    This class contains methods to find yum updates and print
    information of RPM updates.
    It uses yum python API client to find the updates.
```

Removed containers as this can also work on normal systems.

Dockerfiles/ccp-openshift-scan/scanning/yumupdates.py
navidshaikh: ok.

dharmit:
LGTM, waiting for CI.

PR #630, opened 2018-09-19
Adding openshift cmd client and moving utils to appropriate location.

mohammedzee1000:
- Creating basic structure for clients
 - Moving what are supposed to be utils to utils package.
 - Modifying index reader accordingly.
 - Creating OpenshiftCmdClient.


ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,246 @@
...
+    Client for interacting with Openshift.
+    """
+
+    def __init__(self):
+        super(OpenShiftCmdClient, self).__init__(
+            "/usr/bin/oc"
+        )
+
+    @retry(tries=10, delay=3, backoff=2)
+    def get_sa_token_from_openshift(self, namespace, sa="sa/jenkins"):
```
dharmit: There's a similar code at https://github.com/CentOS/container-pipeline-service/blob/ef004815ef119f8619ed91657de3dbf1aca15a49/ccp/lib/openshift.py#L14-L31

Can we reuse/modify that and get things done instead?

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,246 @@
...
+        :param token: Default None: The token to login with. Overrides username
+        and password as preferred login method
+        :type token str
+        :param username: Default none : The username to login with. Use only if
+        token is not used.
+        :type username str
+        :param password: The password to login with. Use only if token is not
+        provided. Not recommended
+        :type password str
+        :return: The output ot the command
```
dharmit: *output of the command

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,246 @@
...
+        :rtype str
+        :raises Exception
+        """
+        command = str.format(
+            "{base_command} login{token_param}{user_param}{server_param}",
```
dharmit: Should there be a space between `login` and `{token_param}`?

dharmit: Similarly should there be a space between each of `{token_param}`, `{user_param}` and `{server_param}`?

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,246 @@
...
+            user_param="" if token else " -u {} -p {}".format(
+                username,
+                password
+            ),
+            server_param="" if not server else str(server)
+        )
+        return self.run_command(cmd=command)
+
+    @retry(tries=10, delay=3, backoff=2)
+    def process_template(
```
dharmit: Out of curiosity, where do we need this function? Can you cite an example? Because processing a template would mean we make changes to the underlying cluster. However, AFAIK, we're trying to fetch the info from Jenkins and OpenShift as a part of this effort.

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,246 @@
...
+        :param namespace: The namespace in which the template is to be applied
+        :param apply_template: Default True: Apply the template
+        :type template_path str
+        :type params dict
+        :type namespace str
+        :type apply_template bool
+        :returns output of command execution
+        :rtype str
+        :raises Exception
+        """
```
dharmit: Should we add a check here to verify if a file exists at given `template_path` or not?

ccp/lib/clients/base.py
```diff
@@ -0,0 +1,42 @@
...
+        :param cmd: The command to execute
+        :type cmd str
+        :param shell: Default False: Enable advanced shell function, not
+        recommended
+        :type shell bool
+        :return:The output of command
+        :rtype str
+        :raises Exception
+        """
+        out, err = run_cmd2(cmd, shell=shell)
```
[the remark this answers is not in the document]
mohammedzee1000: This is for cases when there is no exception, but just an error message. This would be a base class, and hence need to be as generic as possible.

ccp/lib/clients/openshift/client.py
mohammedzee1000: I have kept the file so as to not break anything. Once index reader is refactored, this file can be dropped.

ccp/lib/clients/openshift/client.py
mohammedzee1000: Yes, there will be, if any of them are set, they add space themselves. Otherwise, it is just an extra character

ccp/lib/clients/openshift/client.py
mohammedzee1000: In the index reader, where we have explicit strings scattered across it for applying seed job and weekly scan job etc. Atleast that is what i remember off the top of my head.

mohammedzee1000: This PR does include stuff that is not immediate requirement. But can be used as we go forward, post the API piece. I added is as i did not have anything else to do anyway. And as already said, we should do anyway.

Also, this does not conflict with the title of PR anycase.

ccp/lib/clients/base.py
```diff
@@ -0,0 +1,42 @@
...
+class Client(object):
+    """
+    Base class for all clients
+    """
+    pass
+
+
+class CmdClient(Client):
+    """
+    Base class for all Command Clients
```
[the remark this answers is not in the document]
mohammedzee1000: Good point adding more description.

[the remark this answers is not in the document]
mohammedzee1000: Added

ccp/lib/clients/base.py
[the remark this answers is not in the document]
bamachrn: why we are using two run commands? one with check output and another with Popen, may be we should change the name of the function from `run_cmd2` to `run_cmd_popen` or something.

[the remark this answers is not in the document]
bamachrn: here also we are rewriting the function, why dont we just use the same `run_cmd` or `run_cmd2` from library?

ccp/lib/clients/openshift/client.py
bamachrn: agree with @dharmit, that makes code better maintainable

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,258 @@
...
+        )
+        return self.run_command(command, shell=True)
+
+    @retry(tries=5, delay=3, backoff=3)
+    def get_token_from_mounted_secret(
+            self, secret_mount_path="/tmp/jenkins-secret"
+    ):
+        """
+        Get token from a mounted secret.
+        :param secret_mount_path: The path where the secret is mounted
```
bamachrn: this is mount path from slave container, not path from the /etc/origin right?

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,258 @@
...
+        :return: The token, if it was able to retrieve it.
+        """
+        token_location = secret_mount_path + "/token"
+        if not path.exists(token_location):
+            raise Exception(
+                "Could not find token at {}".format(
```
bamachrn: we are raising all the exceptions in the function, lets make sure we are actually caching them properly. Program should not stop working as expected, even though there is exception

ccp/lib/clients/openshift/client.py
bamachrn: dont we need configs for certs and other details for login?

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,258 @@
...
+        if not filter_out or not isinstance(filter_out, list):
+            filter_out = []
+        selector_params = ""
+        for k, v in selectors:
+            selector_params = selector_params + " --selector {k}={v}".format(
+                k=k,
+                v=v
+            )
+        command = "{base_command} get bc -o name -n {namespace}{s}".format(
+            base_command=self.base_command,
```
bamachrn: so this base command takes care of specifying `oc` or `oc --config <config>` right?

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,258 @@
...
+            namespace=namespace,
+            s=selector_params
+        )
+        out = self.run_command(cmd=command)
+        if not out.strip():
+            return []
+        out = out.strip().split('\n')
+        return [each for each in out if each not in filter_out]
```
bamachrn: got puzzled here, this returns a list or string? also can we use some variable name instead of `each`?

ccp/lib/clients/openshift/client.py
dharmit: I request to do the necessary refactoring in this iteration itself. If something's not looking good, let's talk and fix it. Because piling things up is going to lead us to a round of refactoring that will require all of us to work on it for a sprint or two.

mohammedzee1000: OK,  I will get that done then

ccp/lib/clients/openshift/client.py
mohammedzee1000: Yes.  It will be the path where the secret is mounted

ccp/lib/clients/openshift/client.py
mohammedzee1000: Yes.  But it should be handled at higher level the way they want to handle it.

ccp/lib/clients/openshift/client.py
mohammedzee1000: Yes

ccp/lib/clients/openshift/client.py
mohammedzee1000: Yes.  This is list comprehension. It will return a list of items.  This is basically form a list from out.  For each item in out.  If that item is not in filter_out which is another list.  It will be added to the list that will be returned

mohammedzee1000: If u remember mathematical notations from set theory, this is equivalent to
{each:each € out and each € with / filter out }

ccp/lib/clients/base.py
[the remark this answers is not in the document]
mohammedzee1000: Moved to single function to for all run_command operations.

ccp/lib/clients/openshift/client.py
mohammedzee1000: Done. For now i have updated it to use OpenshiftCmdClient to get the jenkins token.. 

Note the file will be removed after the Openshift Jenkins Processor is merged. That bit is outside the scope of this PR.

ccp/lib/clients/openshift/client.py
mohammedzee1000: Oh yea or atleast insecure-skip if nothing else.  I will add the same

ccp/lib/clients/openshift/client.py
mohammedzee1000: Added

ccp/lib/clients/openshift/client.py
mohammedzee1000: FYI exceptions should rarely be handled at lower level of code. Unless there is explicit need. This should be done at higher level as each client of lib code will want to handle exceptions in different way. Not even processor. 

What will end up happening is API server for example will be the highest level that needs to know there is exception and handle it correctly.

Another example will be index reader. It should be the one that decides what action it wants to take it when an oc command fails for example. The duty of lower level code that runs the command is updating the reader of the condition.

mohammedzee1000:
@bamachrn the error seems to be on your side. Maybe you can see what happened here

bamachrn:
Seems like we did not get the nodes from duffy pool

bamachrn:
#dotests

bamachrn:
We need to update the CI, oc process command for the seed-job

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        if not filter_out:
+            filter_out = []
+
+        if status_filter:
+            conditional = '(ne .status.phase "{}") '
+            condition = ''
+            for phase in status_filter:
+                condition = condition + conditional.format(phase)
+            filter_str = "{{if and %s }}" % condition
+            close_str = close_str * 2
```
navidshaikh: if you are adding the `close_str` not in any loop, why to add an extra operation here? Just define `{{end}} {{end}}` at once?

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        :type status_filter list
+        :type filter_out list
+        :return: List of builds that matched all conditions
+        :rtype list
+        :raises Exception
+        :raises subprocess.CalledProcessError
+        :raises ccp.lib.exceptions.CommandOutputError
+        """
+        filter_str = ""
+        close_str = "{{end}}"
```
navidshaikh: `filter_str` and `close_str` only applies `if status_filter` - so define these two variables only `if status_file`  is true.

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+            condition = ''
+            for phase in status_filter:
+                condition = condition + conditional.format(phase)
+            filter_str = "{{if and %s }}" % condition
+            close_str = close_str * 2
+
+        template_str = "{}{}{}".format(
+            "{{range .items }}",
+            filter_str,
+            "{{.metadata.name}}:{{.status.phase}} " + close_str
```
navidshaikh: this will put 3 `close_str` at the end while only two are needed

also just ensuring, the format doesn't replace two `{` ---to---> single `{` for `{{range .items }}` right ?
(just use ipython to ensure this)

ccp/lib/email.py
```diff
@@ -54,7 +52,7 @@ def email(self,
             to_addresses=to_addresses)
 
         # send email
-        stdout, stderr = run_cmd2(command, shell=True)
+        stdout, stderr = run_command(command, shell=True)
```
navidshaikh: This is major breakage!

The `run_cmd2` is supposed to return `stdout and stderr`. You have updated the `run_command` to return only one value.

There are lines in code where run_cmd is used differently, at places only one value is expected while 2 at some other places.
if you are refactoring that, I'd suggest we use the `stdout` and `stderr` approach and update all the references of `run_command` to work with two values.

ccp/lib/exceptions.py
```diff
@@ -11,6 +11,13 @@ class InvalidPipelineName(Exception):
     pass
 
 
+class CommandOutputError(Exception):
+    """
+    Exception to be raised when running a command does not fail but command
+    gives an error message.
+    """
```
navidshaikh: add the keyword `pass`

ccp/lib/openshift.py
```diff
@@ -202,7 +183,9 @@ def get_build_info(
         print ("Opening URL to details of the build\n{}".format(url))
 
         try:
-            token = get_jenkins_access_token(self.service_account)
+            token = OpenShiftCmdClient().get_sa_token_from_openshift(
```
navidshaikh: not initializing it in the __inif__ of `BuildInfo` class ? I'd say we do that, initialize in the __init__ and reference via an object.

ccp/lib/utils/command.py
```diff
@@ -0,0 +1,20 @@
...
+    :param shell: Default False: Whether to run raw shell commands with '|' and
+    redirections
+    :return: Command output
+    :raises subprocess.CalledProcessError
+    :raises ccp.lib.exceptions.CommandOutputError
+    """
+    p = subprocess.Popen(cmd, shell=shell)
+    out, err = p.communicate()
+    if err:
+        raise CommandOutputError(err)
```
navidshaikh: Let the caller decide what needs to be done if `err` comes - and let the caller handle the exceptions.

Also, this breaks the functionality where `run_cmd` and `run_cmd2` are used, expecting `1` and `2` return values respectively.

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        client_cert_path = client_cert_path or self.secure_client_cert_path
+        client_key_path = client_key_path or self.secure_client_key_path
+        cert_params = "{}{}{}".format(
+            "" if not ca_path else "--certificate-authority=" + ca_path,
+            "" if not client_cert_path else "--client-certificate=" +
+                                            client_cert_path,
+            "" if not client_key_path else "--client-key=" + client_key_path
+        )
+        command = str.format(
+            "{base_command}{secure_params} login{token_param}{user_param}"
```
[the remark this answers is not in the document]
mohammedzee1000: These spaces are added by the parameters themselves. For example for params the for loop add " -p blah=blah" on its own. Similarly secure_params will add its own space if it is adding anything. Else it equals ""

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+                )
+            )
+        p = ""
+        for k, v in params:
+            p = p + " -p {param_name}={param_value}".format(
+                param_name=k,
+                param_value=v
+            )
+        command = str.format(
+            "{base_command} process{params} -f {template_path}{apply_cmd}",
```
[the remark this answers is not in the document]
mohammedzee1000: As stated. This will be added if there are params to add. It is an optional param. If no params are passed. There will then be an unneeded space.

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        if not filter_out or not isinstance(filter_out, list):
+            filter_out = []
+        selector_params = ""
+        for k, v in selectors:
+            selector_params = selector_params + " --selector {k}={v}".format(
+                k=k,
+                v=v
+            )
+        command = "{base_command} get bc -o name -n {namespace}{s}".format(
```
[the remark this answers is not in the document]
mohammedzee1000: I am not putting spaces for optional params. These are going to add their own space, if they are being set

ccp/lib/email.py
mohammedzee1000: Ack fixed

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+
+    ):
+        """
+        Initialize OpenshiftCmdClient
+        :param base_command_args: Any special args you wish to give to base
+        command
+        :type base_command_args list
+        :param insecure: Default True: If true, then connection is insecure even
+        with https so certs verification is ignored
+        :type insecure bool
```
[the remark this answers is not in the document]
mohammedzee1000: done

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+    """
+
+    def __init__(
+            self,
+            base_command_args=None,
+            insecure=True,
+            ca_path=None,
+            client_cert_path=None,
+            client_key_path=None
+
```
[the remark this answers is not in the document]
mohammedzee1000: Removed

ccp/lib/clients/openshift/client.py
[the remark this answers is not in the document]
navidshaikh: Nope, that's not right. We need to define the spaces in the formula of string formatting itself, and then we can rest assure if the parameters' value are adding spaces or not. This way we follow more explicit approach and leave no room for confusion and error.

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        :type ca_path str
+        :param client_cert_path: Default None: If provided then client cert path
+        is used.
+        :type client_cert_path str
+        :param client_key_path: Default None: If provided, then client key path
+        is used.
+        :type client_key_path str
+        """
+        super(OpenShiftCmdClient, self).__init__(
+            "/usr/bin/oc{}".format(
```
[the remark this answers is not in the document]
mohammedzee1000: The space is appended if base_command_args are provided

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        :return: The token, if it was able to retrieve it.
+        """
+        token_location = secret_mount_path + "/token"
+        if not path.exists(token_location):
+            raise Exception(
+                "Could not find token at {}".format(
+                    secret_mount_path
+                )
+            )
+        with open(token_location, "r") as f:
```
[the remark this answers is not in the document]
mohammedzee1000: Got that but i'd still rather be explicit than rely on defaults. As much as possible

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+            base_command=self.base_command,
+            secure_params="{insecure_param}{cert_params}".format(
+                insecure_param="--insecure-skip-tls-verify=true" if insecure
+                else "",
+                cert_params=cert_params
+            ),
+            token_param="" if not token else " --token={}".format(
+                token
+            ),
+            user_param="" if token else " -u {} -p {}".format(
```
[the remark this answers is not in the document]
mohammedzee1000: Sorted

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        :type apply_template bool
+        :returns output of executed command
+        :rtype str
+        :raises Exception
+        :raises subprocess.CalledProcessError
+        :raises ccp.lib.exceptions.CommandOutputError
+        """
+        if not path.exists(template_path):
+            raise Exception(
+                "Invalid path of template file at {}".format(
```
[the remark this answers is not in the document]
mohammedzee1000: Updated to better format

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+                    template_path
+                )
+            )
+        p = ""
+        for k, v in params:
+            p = p + " -p {param_name}={param_value}".format(
+                param_name=k,
+                param_value=v
+            )
```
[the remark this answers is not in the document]
mohammedzee1000: Changed

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        command = str.format(
+            "{base_command} process{params} -f {template_path}{apply_cmd}",
+            base_command=self.base_command,
+            params=p,
+            template_path=template_path,
+            apply_cmd=" | {base_command} apply -n {ns} -f -".format(
```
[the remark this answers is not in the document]
mohammedzee1000: Fixed

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+        :raises subprocess.CalledProcessError
+        :raises ccp.lib.exceptions.CommandOutputError
+        """
+        if not filter_out or not isinstance(filter_out, list):
+            filter_out = []
+        selector_params = ""
+        for k, v in selectors:
+            selector_params = selector_params + " --selector {k}={v}".format(
+                k=k,
+                v=v
```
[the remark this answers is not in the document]
mohammedzee1000: Changed

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,342 @@
...
+            )
+        command = "{base_command} get bc -o name -n {namespace}{s}".format(
+            base_command=self.base_command,
+            namespace=namespace,
+            s=selector_params
+        )
+        out = run_command(cmd=command)
+        if not out.strip():
+            return []
+        out = out.strip().split('\n')
```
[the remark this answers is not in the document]
mohammedzee1000: That wont be the case with this command or any oc command where output format is provided with -o or --template

ccp/lib/exceptions.py
mohammedzee1000: Add it where here?

ccp/lib/clients/openshift/client.py
mohammedzee1000: Nope. filter_str is the only one what applies in case of status_filter. close_str applies either way. It is just number of instances of it that change.

ccp/lib/clients/openshift/client.py
mohammedzee1000: I have removed multiply in favour of re-initialisation

ccp/lib/clients/openshift/client.py
mohammedzee1000: I know it looks more efficient. But it is simply easier to keep it as is.

ccp/lib/utils/command.py
mohammedzee1000: Done

ccp/lib/exceptions.py
navidshaikh: add in the empty class definition

ccp/lib/clients/openshift/client.py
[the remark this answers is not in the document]
navidshaikh: Nope @mohammedzee1000  - lets add the spaces in hereitself and not rely on the variable's value population, this is prone to errors - also think about this, you need to add space only once, and not worry about adding space in every variable populated later.

ccp/lib/clients/openshift/client.py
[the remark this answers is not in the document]
navidshaikh: the unneeded space doesn't through an error in bash commands,  no-space does.

ccp/lib/clients/openshift/client.py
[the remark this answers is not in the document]
navidshaikh: as stated above, extra spaces wont through error, but if param's value processing doesn't have an space, it will.

ccp/lib/clients/openshift/client.py
```diff
@@ -0,0 +1,344 @@
...
+        if not filter_out:
+            filter_out = []
+
+        if status_filter:
+            conditional = '(ne .status.phase "{}") '
+            condition = ''
+            for phase in status_filter:
+                condition = condition + conditional.format(phase)
+            filter_str = "{{if and %s }}" % condition
+            close_str = "{{end}}{{end}}"
```
navidshaikh: repeated definitions of `close_str` first occurrence at line 314 in same file

ccp/lib/utils/command.py
```diff
@@ -0,0 +1,42 @@
...
+    :param shell: Default False: Whether to run raw shell commands with '|' and
+    redirections
+    :param use_pipes: Default False: If true, then pipes are used for
+    communication
+    :return: Command output and err
+    :raises subprocess.CalledProcessError
+    """
+    p = subprocess.Popen(
+        cmd, shell=shell, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    ) if use_pipes else subprocess.Popen(
```
navidshaikh: just make sure, the command which is going to return a lot of data (for eg, logs), we don't use the pipes as, it may chock up the available memory and block the command.

navidshaikh: I'd default on not using the pipe and when needed specify that explicitly.

ccp/lib/clients/openshift/client.py
mohammedzee1000: Yes. It is meant to be overridden. If filter_out is provided then close_str="{{end}}{{end}}". Otherwise close_str="{{end}}". 

That is why it is being set first to one occurrence of {{end}}, and if filter_out is provided then it is reinitialised to 2 occurrences.

ccp/lib/utils/command.py
mohammedzee1000: In most cases, i would agree with you. But here it is better this way.

In most usages in our code, we want to pipe the output. It is rare case where we just want to execute a command and dont care what it gives as output. That is why default is True.

It is for program efficiency.

PR #634, opened 2018-09-20
Adds unittests and functional CI debug jobs

bamachrn:
* this enables #dotests-debug for keeping the duffy nodes once  the CI is over.
This helps us debug further on the issues if required.


ci/job.yml
```diff
@@ -113,15 +113,79 @@
...
+
+- job:
+    name: 'centos-container-pipeline-service-ci-pr-functional-debug'
+    description: |
+        Managed by Jenkins Job Builder, do not edit manually!
+    node: container
+    ci_project: container
+    properties:
+        - github:
+            url: https://github.com/CentOS/container-pipeline-service/
```
navidshaikh: We can have git URL specified in same way throughout the job like
```
            url: https://github.com/CentOS/container-pipeline-service.git
```
specified [here](https://github.com/CentOS/container-pipeline-service/pull/634/files#diff-091765ca1f913cb1fc03c4831eb431bbR166) at line no 165 later in the file.

ci/job.yml
```diff
@@ -113,15 +113,79 @@
...
+        - shell: |
+            echo ${sha1}
+            echo ${GIT_URL}
+            echo ${GIT_BRANCH}
+            CI_DEBUG=1
+            if [ ${ghprbTargetBranch} == "openshift" ]
+            then
+              git rebase origin/${ghprbTargetBranch}
+              cd ci
+              sh trigger_ci.sh ${GIT_URL} ${GIT_BRANCH} ${sha1} ${CI_DEBUG}
```
navidshaikh: `sh ci/trigger_ci.sh [..]` is fine too, no need to cd into that directory.

ci/trigger_ci.sh
```diff
@@ -5,7 +5,14 @@ mark_failure()
     echo "$1"
     echo "=================================================="
     echo "CI complete releasing the nodes"
-    cico node done $cico_node_key
+    if [ $CI_DEBUG -eq 0]
+    then
+        cico node done $cico_node_key
+    else
+        echo "============================================================"
+        echo "DEBUG mode is set for CI, keeping nodes for debugging"
```
navidshaikh: you need to define a debug period here, for example sleep for M seconds for debug, after lapsing of M seconds, we need to release the nodes.

This is to ensure the nodes are not left hanging behind if not debugged within specified time.

We had this behavior earlier.

navidshaikh:
I will be pushing required changes to this PR.

ci/job.yml
```diff
@@ -113,15 +113,79 @@
...
+                - cdrage
+            white-list:
+                - kbsingh
+                - bamachrn
+                - dharmit
+                - navidshaikh
+                - mohammedzee1000
+                - cdrage
+            trigger-phrase: '#dotests-debug'
+            only-trigger-phrase: false
```
navidshaikh: @bamachrn : We need to set this to `true`, so that, this debug job doesn't trigger on every PR but only when it finds the comment as defined.

PR #636, opened 2018-09-21
Stops triggering builds for configChange of master-job template

navidshaikh:
All the pipelines have master-job template set for building at OpenShift.

 Any single change in template, is identified as configChange by oc.

 In IndexReader, we see if a buildConfig / master-job for pipeline is configured,
 while running seed-job (which syncs index with pipelines on OpenShift).

 Update in tempaltes can be of two types:
  A- Update in parameters which affects the resultant builds
  B- Update in parameters which affects service side workflow and behavior

 The build should only be re-triggered in case A.

 However, those changes as mentioned in A, are already taken
 care by poll SCM config in master-job of pipeline.

 Thus, we need to prevent re-triggering builds in case of B.

### DIFF
--- a/ccp/index_reader.py
+++ b/ccp/index_reader.py
@@ -309,11 +309,20 @@ class BuildConfigManager(object):
         # if a buildConfig has config update, oc apply returns
         # "buildconfig.build.openshift.io "$PIPELINE_NAME" configured"
         # possible values are ["unchanged", "created", "configured"]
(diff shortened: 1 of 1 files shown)


navidshaikh:
#dotests

bamachrn:
#dotests

bamachrn:
#dotests

PR #637, opened 2018-09-24
Setting up the Openshift Jenkins Client

mohammedzee1000:
This PR sets up the OpenshiftJenkinsClient to allow easy communication with jenkins on openshift.

### DIFF
--- a/ccp/lib/clients/base.py
+++ b/ccp/lib/clients/base.py
@@ -1,4 +1,6 @@
 from ccp.lib.utils.command import run_command
+from ccp.lib.utils.request import request_url
+from ccp.lib.utils.authorization import Authorization
 
 
 class Client(object):
@@ -23,3 +25,57 @@ class CmdClient(Client):
         """
         super(Client, self).__init__()
         self.base_command = base_command
+
+
+class APIClient(Client):
+    """
+    Base class for all clients that query an API Server for information
+    """
+    def __init__(
+            self,
+            server,
+            secure=True,
+            verify_ssl=True,
+            authorization=None
+    ):
+        """
+        Initialize APIClient
+        :param server: The URL/IP server to hit.
+        :type server str
+        :param secure: Default True: Use SSL for queries.
+        :type secure bool
+        :param verify_ssl: Default True: Verify SSL certificate.
+        :param authorization: Set if you want or need to use Authorization
+        :type authorization Authorization
+        """
+        self.server_endpoint = "{}://{}".format(
+            "https" if secure else "http",
+            server if server else "localhost"
+        )
+        self.authorization = authorization
+        self.skip_ssl = verify_ssl
+
+    def _query(self, target, headers=None):
+        """
+        Given a target (query without url), queries endpoint and gets data
+        :param target: The query to send to the api server, after the hostname
+        and port. For eg: URL = http://example.com/hello, then target will be
+        '/hello'
+        :type target str
+        :param headers: Any headers you wish to add to query.
+        :type headers dict
+        :return:The response object and exception object, if any was raised.
+        """
+        query_url = "{}{}".format(
+            self.server_endpoint,
+            target
+        )
+        request_headers = {}
+        if self.authorization and isinstance(self.authorization, Authorization):
+            self.authorization.add_header(request_headers)
+        if headers and isinstance(headers, dict):
+            request_headers.update(headers)
+        return request_url(
+            query_url, verify_ssl=self.skip_ssl, headers=request_headers
+        )
+
--- a/ccp/lib/clients/base.py
+++ b/ccp/lib/clients/base.py
@@ -1,6 +1,9 @@
-from ccp.lib.utils.command import run_command
+"""
+This file contains all the base classes for clients
+"""
 from ccp.lib.utils.request import request_url
 from ccp.lib.utils.authorization import Authorization
+from urlparse import urljoin
 
 
 class Client(object):
@@ -19,7 +22,7 @@ class CmdClient(Client):
 
     def __init__(self, base_command):
         """
-        Iniitalize the Cmd Client
+        Initialize the Cmd Client
         :param base_command: The core command to be used by client
         :type base_command: str
         """
@@ -40,13 +43,13 @@ class APIClient(Client):
     ):
         """
         Initialize APIClient
-        :param server: The URL/IP server to hit.
-        :type server str
-        :param secure: Default True: Use SSL for queries.
+        :param server: The URL/IP server to hit
+        :type server: str
+        :param secure: Default True: Use SSL for queries
         :type secure bool
-        :param verify_ssl: Default True: Verify SSL certificate.
+        :param verify_ssl: Default True: Verify SSL certificate
         :param authorization: Set if you want or need to use Authorization
-        :type authorization Authorization
+        :type authorization: Authorization
         """
         self.server_endpoint = "{}://{}".format(
             "https" if secure else "http",
@@ -61,15 +64,12 @@ class APIClient(Client):
         :param target: The query to send to the api server, after the hostname
         and port. For eg: URL = http://example.com/hello, then target will be
         '/hello'
-        :type target str
+        :type target: str
         :param headers: Any headers you wish to add to query.
-        :type headers dict
-        :return:The response object and exception object, if any was raised.
+        :type headers: dict
+        :return: The response object and exception object, if any was raised.
         """
-        query_url = "{}{}".format(
-            self.server_endpoint,
-            target
-        )
+        query_url = urljoin(self.server_endpoint, target)
         request_headers = {}
         if self.authorization and isinstance(self.authorization, Authorization):
             self.authorization.add_header(request_headers)
--- a/ccp/lib/clients/base.py
+++ b/ccp/lib/clients/base.py
@@ -47,7 +47,10 @@ class APIClient(Client):
         :type server: str
         :param secure: Default True: Use SSL for queries
         :type secure bool
-        :param verify_ssl: Default True: Verify SSL certificate
+        :param verify_ssl: Default True: Verify SSL certificate. You can also
+        provide path of cert to verify with, if your vert resides outside
+        default location
+        :type verify_ssl: Union[bool, str]
         :param authorization: Set if you want or need to use Authorization
         :type authorization: Authorization
         """
@@ -56,7 +59,7 @@ class APIClient(Client):
             server if server else "localhost"
         )
         self.authorization = authorization
-        self.skip_ssl = verify_ssl
+        self.verify_ssl = verify_ssl
 
     def _query(self, target, headers=None):
         """
@@ -76,6 +79,6 @@ class APIClient(Client):
         if headers and isinstance(headers, dict):
             request_headers.update(headers)
         return request_url(
-            query_url, verify_ssl=self.skip_ssl, headers=request_headers
+            query_url, verify_ssl=self.verify_ssl, headers=request_headers
         )
 
--- a/ccp/lib/clients/jenkins/base.py
+++ b/ccp/lib/clients/jenkins/base.py
@@ -27,6 +27,7 @@ class OpenshiftJenkinsBaseAPIClient(APIClient):
             server,
             secure=True,
             verify_ssl=True,
+            token=None,
             token_from_mount=None,
             sa="sa/jenkins",
             namespace="default"
@@ -38,8 +39,13 @@ class OpenshiftJenkinsBaseAPIClient(APIClient):
         :param secure: Default True: Use SSL for queries.
         :type secure bool
         :param verify_ssl: Default True: Verify SSL certificate.
+        :type verify_ssl bool
+        :param token: Default None: If provided then, this is set as the token
+        to use to login to OpenShift. Overrides all other ways of providing
+        token
+        :type token str
         :param token_from_mount: Default None: Set if you have token mounted
-        at a path. Otherwise, ensure the openshift context is already set.
+        at a path. Otherwise, ensure the OpenShift context is already set.
         :type token_from_mount str
         :param sa: Default 'sa/jenkins': Name of the service account whose
         token is to be used.
@@ -48,11 +54,12 @@ class OpenshiftJenkinsBaseAPIClient(APIClient):
         :type namespace str
         """
 
-        c = OpenShiftCmdClient()
-        token = c.get_token_from_mounted_secret(token_from_mount)\
-            if token_from_mount else c.get_sa_token_from_openshift(
-            sa=sa, namespace=namespace
-        )
+        if not token:
+            c = OpenShiftCmdClient()
+            token = c.get_token_from_mounted_secret(token_from_mount)\
+                if token_from_mount else c.get_sa_token_from_openshift(
+                sa=sa, namespace=namespace
+            )
         super(OpenshiftJenkinsBaseAPIClient, self).__init__(
             server=server,
             secure=secure,
--- a/ccp/lib/clients/jenkins/base.py
+++ b/ccp/lib/clients/jenkins/base.py
@@ -8,7 +8,7 @@ def jenkins_jobs_from_jobs_ordered_list(nested_job_ordered_list):
     Formats query part of URL from ordered job list as /job/j1/job/j2
     :param nested_job_ordered_list: The ordered list of jenkins job names,
     with parent as first and every child below.
-    :type nested_job_ordered_list list
+    :type nested_job_ordered_list: list
     :return: A string of the form /job/j1/job/j2...
     """
     dest = ""
@@ -35,23 +35,23 @@ class OpenshiftJenkinsBaseAPIClient(APIClient):
         """
         Initialize Jenkins APIClient
         :param server: The URL/IP server to hit.
-        :type server str
+        :type server: str
         :param secure: Default True: Use SSL for queries.
-        :type secure bool
+        :type secure: bool
         :param verify_ssl: Default True: Verify SSL certificate.
-        :type verify_ssl bool
+        :type verify_ssl: bool
         :param token: Default None: If provided then, this is set as the token
         to use to login to OpenShift. Overrides all other ways of providing
         token
-        :type token str
+        :type token: str
         :param token_from_mount: Default None: Set if you have token mounted
         at a path. Otherwise, ensure the OpenShift context is already set.
-        :type token_from_mount str
+        :type token_from_mount: str
         :param sa: Default 'sa/jenkins': Name of the service account whose
         token is to be used.
-        :type sa str
+        :type sa: str
         :param namespace: The namespace of the Jenkins secret, if not mounted
-        :type namespace str
+        :type namespace: str
         """
 
         if not token:
--- a/ccp/lib/clients/jenkins/base.py
+++ b/ccp/lib/clients/jenkins/base.py
@@ -1,0 +1,62 @@
+from ccp.lib.clients.base import APIClient
+from ccp.lib.clients.openshift.client import OpenShiftCmdClient
+from ccp.lib.utils.authorization import Authorization, BearerAuthorization
+
+
+def jenkins_jobs_from_jobs_ordered_list(nested_job_ordered_list):
+    """
+    Formats query part of URL from ordered job list as /job/j1/job/j2
+    :param nested_job_ordered_list: The ordered list of jenkins job names,
+    with parent as first and every child below.
+    :type nested_job_ordered_list list
+    :return: A string of the form /job/j1/job/j2...
+    """
+    dest = ""
+    for i in nested_job_ordered_list:
+        dest = "{}/job/{}".format(
+            dest,
+            i
+        )
+    return dest
+
+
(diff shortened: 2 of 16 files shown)


mohammedzee1000:
Relies on https://github.com/CentOS/container-pipeline-service/pull/630

ccp/lib/utils/request.py
```diff
@@ -0,0 +1,29 @@
+import requests
+
+
+def request_url(
+        request, params=None, verify_ssl=False, auth=None, headers=None
+):
+    """
+    Queries a specified URL and returns data, if any
+    :param request: The url to send the request to.
+    :type request str
```
navidshaikh: `:type request: str`

navidshaikh: applies to all occurrences where `type` is defined.

ccp/lib/utils/request.py
```diff
@@ -0,0 +1,29 @@
...
+    :param params: Any params that are to be passed on to the request
+    :type params dict
+    :param verify_ssl: IF False, ssl errors are ignored.
+    :type verify_ssl bool
+    :param auth: This will be passed along to requests as is. Note: Passing
+    auth overrides any Authorization headers passed already.
+    :type auth dict
+    :param headers: Any extra headers that you wish to pass along.
+    :type headers dict
+    :raises Exception
```
navidshaikh: What kind of Exception it raises ?

ccp/lib/utils/request.py
```diff
@@ -0,0 +1,29 @@
+import requests
```
navidshaikh: It would be good, if we can add one-line at the start of the file for eg
`"This file handles API requests and returns the response for given API URL along with provided parameters. "`

navidshaikh: applies to all other files as well.

ccp/lib/utils/parsing.py
```diff
@@ -0,0 +1,55 @@
+import ast
+import base64
+import json
+
+
+def encode(data):
+    """
+    Applyies base64 encoding to passed string
```
navidshaikh: typo, `Applies`

ccp/lib/utils/parsing.py
```diff
@@ -0,0 +1,55 @@
...
+    :param data: The string to encode
+    :type data str
```
navidshaikh: `:type data: str`

ccp/lib/utils/parsing.py
```diff
@@ -0,0 +1,55 @@
...
+    :return: Encoded string.
```
navidshaikh: remove the unneeded dot (.) at the end of `Encoded string.`

ccp/lib/utils/parsing.py
```diff
@@ -0,0 +1,55 @@
...
+    :param data: The json data to parse as string
+    :type data str
+    :raises Exception
+    :return: Pythonic repersentation of json data
+    """
+    p = None
+    try:
+        p = json.loads(data)
+    except Exception as ex:
+        raise ex
```
navidshaikh: if you are raising the same exception as occurred, its same as
`return json.loads(data)`

as if, the exception is raised, `p = None` is never used.

ccp/lib/utils/parsing.py
```diff
@@ -0,0 +1,55 @@
...
+    quotes.
+    :param data The data to be parsed as literal as a string.
+    :type data str
+    :raises Exception
+    :return: Pythonic repersentation of data
+    """
+    p = None
+    try:
+        p = ast.literal_eval(data)
+    except Exception as ex:
```
navidshaikh: same as above comment, its same as
```
return ast.literal_eval(data)
```

ccp/lib/utils/authorization.py
```diff
@@ -0,0 +1,24 @@
+"""
+This file contains authorization handlers
+"""
```
navidshaikh: :+1:  for adding the helper docs at the top of file.

ccp/lib/utils/authorization.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Authorization(object):
```
navidshaikh: We need the class docs as well.

ccp/lib/utils/authorization.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def __init__(self):
```
navidshaikh: shouldn't you be expecting some constructor params here?

navidshaikh: that you will add to `self.authorization_header_string` var ?

ccp/lib/utils/authorization.py
```diff
@@ -0,0 +1,24 @@
...
+        self.authorization_header_string = ""
+
+    def add_header(self, headers):
+        """
+        Adds appropriate authorization headers to request headers
+        :param headers: The headers dict
+        :type headers dict
+        """
+        headers["Authorization"] = self.authorization_header_string
```
navidshaikh: is this always expected to be `""`?

ccp/lib/utils/authorization.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class BearerAuthorization(Authorization):
+
+    def __init__(self, token):
+        super(BearerAuthorization, self).__init__()
```
navidshaikh: I would initialize the parent class constructor with value `"Bearer %s" % token` instead of overriding the variable of parent class.

ccp/lib/clients/jenkins/client.py
```diff
@@ -0,0 +1,127 @@
+"""
+This file contains ready to use Jenkins Clients.
```
navidshaikh: :+1:

ccp/lib/clients/base.py
```diff
@@ -23,3 +25,57 @@ def __init__(self, base_command):
...
+            self,
+            server,
+            secure=True,
+            verify_ssl=True,
+            authorization=None
+    ):
+        """
+        Initialize APIClient
+        :param server: The URL/IP server to hit.
+        :type server str
```
navidshaikh:
```
:type server: str
```

ccp/lib/clients/base.py
```diff
@@ -23,3 +25,57 @@ def __init__(self, base_command):
...
+    def __init__(
+            self,
+            server,
+            secure=True,
+            verify_ssl=True,
+            authorization=None
+    ):
+        """
+        Initialize APIClient
+        :param server: The URL/IP server to hit.
```
navidshaikh: remove the dot (.) for all the strings which are not sentences

ccp/lib/clients/base.py
```diff
@@ -23,3 +25,57 @@ def __init__(self, base_command):
...
+        :type server str
+        :param secure: Default True: Use SSL for queries.
+        :type secure bool
+        :param verify_ssl: Default True: Verify SSL certificate.
+        :param authorization: Set if you want or need to use Authorization
+        :type authorization Authorization
+        """
+        self.server_endpoint = "{}://{}".format(
+            "https" if secure else "http",
+            server if server else "localhost"
```
navidshaikh: that means if `server=None`, its `"localhost"`

okay!

ccp/lib/clients/base.py
```diff
@@ -23,3 +25,57 @@ def __init__(self, base_command):
...
+    def _query(self, target, headers=None):
+        """
+        Given a target (query without url), queries endpoint and gets data
+        :param target: The query to send to the api server, after the hostname
+        and port. For eg: URL = http://example.com/hello, then target will be
+        '/hello'
+        :type target str
+        :param headers: Any headers you wish to add to query.
+        :type headers dict
+        :return:The response object and exception object, if any was raised.
```
navidshaikh: add space after `:return:`

ccp/lib/clients/base.py
```diff
@@ -23,3 +25,57 @@ def __init__(self, base_command):
...
+        """
+        query_url = "{}{}".format(
```
navidshaikh: mind checking if `target` starts with  `/` OR `self.server_endpoint` ends with `/`.
This is a common problem and that's why library utils like `urljoin` is devised.

Please use `urljoin` here to combine, target with server end point.

I think its a must.

navidshaikh: I'll request change for this.

We are going to have a lot of places in our code where we need to perform this, and lets make sure we do that via `urljoin` and leave no room for any input validation errors!

ccp/lib/clients/base.py
```diff
@@ -23,3 +25,57 @@ def __init__(self, base_command):
...
+            self.server_endpoint,
+            target
+        )
+        request_headers = {}
+        if self.authorization and isinstance(self.authorization, Authorization):
+            self.authorization.add_header(request_headers)
```
navidshaikh: what happens if `self.authorization` is **not** instance of `Authorization` ?

ccp/lib/clients/base.py
```diff
@@ -23,3 +25,57 @@ def __init__(self, base_command):
...
+        if headers and isinstance(headers, dict):
+            request_headers.update(headers)
+        return request_url(
+            query_url, verify_ssl=self.skip_ssl, headers=request_headers
```
navidshaikh: if `self.authorization` is populated in line no 75, we are not passing it here ?

navidshaikh: request_url formula looks like this 
```
def request_url(
        request, params=None, verify_ssl=False, auth=None, headers=None
):
```

I think we missed on sending `auth=self.authorization`?

ccp/lib/clients/jenkins/base.py
```diff
@@ -0,0 +1,62 @@
+from ccp.lib.clients.base import APIClient
```
navidshaikh: multiple base.py, I'd expect one-liner at the top in each file.

ccp/lib/clients/jenkins/base.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    Formats query part of URL from ordered job list as /job/j1/job/j2
+    :param nested_job_ordered_list: The ordered list of jenkins job names,
+    with parent as first and every child below.
+    :type nested_job_ordered_list list
+    :return: A string of the form /job/j1/job/j2...
+    """
+    dest = ""
+    for i in nested_job_ordered_list:
+        dest = "{}/job/{}".format(
```
navidshaikh: for small formatting parameters, having them on single line makes it more readable for eg
```
for i in nested_job_ordered_list:
  dest  = "{}/job/{}".format(dest, i)

```

ccp/lib/clients/jenkins/base.py
```diff
@@ -0,0 +1,62 @@
...
+            dest,
+            i
+        )
+    return dest
+
+
+class OpenshiftJenkinsBaseAPIClient(APIClient):
```
navidshaikh: class docs

ccp/lib/clients/jenkins/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(
+            self,
+            server,
+            secure=True,
+            verify_ssl=True,
+            token_from_mount=None,
+            sa="sa/jenkins",
+            namespace="default"
```
navidshaikh: umm - we should make this is as required parameter IMO.

navidshaikh: as in, without any defaults

navidshaikh: We'are already discussing about spreading the jobs in equal numbers across different namespaces, then there would be need to invoke the constructor with different namespaces, the default will rarely be used I think, its better to have this var definition without a default value.

ccp/lib/clients/jenkins/base.py
```diff
@@ -0,0 +1,62 @@
...
+        token is to be used.
+        :type sa str
+        :param namespace: The namespace of the Jenkins secret, if not mounted
+        :type namespace str
+        """
+
+        c = OpenShiftCmdClient()
+        token = c.get_token_from_mounted_secret(token_from_mount)\
+            if token_from_mount else c.get_sa_token_from_openshift(
+            sa=sa, namespace=namespace
```
navidshaikh: the ordering of params for `get_sa_token_from_openshift`read 
```
    def get_sa_token_from_openshift(self, namespace, sa="sa/jenkins"):
```

even though you have kwargs, I'd expect to have them specified in right order.

Remember I asked to alter the ordering of params in PR 630?

ccp/lib/clients/jenkins/client.py
```diff
@@ -0,0 +1,127 @@
...
+"""
+
+from ccp.lib.clients.jenkins.base import OpenshiftJenkinsBaseAPIClient, \
+    jenkins_jobs_from_jobs_ordered_list
+
+
+class OpenshiftJenkinsWorkflowAPIClient(OpenshiftJenkinsBaseAPIClient):
+    """
+    Helps query the jenkins workflow API endpoints.
```
navidshaikh: *`Jenkins`

ccp/lib/clients/jenkins/client.py
```diff
@@ -0,0 +1,127 @@
...
+            secure=True,
+            verify_ssl=True,
+            token_from_mount=None,
+            sa="sa/jenkins",
+            namespace="default"
+    ):
+        """
+        Initialize Openshift Jenkins Client
+        :param server: The URL/IP of jenkins server on openshift.
+        :type server str
```
navidshaikh: `:type server: str`

ccp/lib/clients/jenkins/client.py
```diff
@@ -0,0 +1,127 @@
...
+                build_number
+            )
+        )
+
+    def describe_execution_node(
+            self, job_ordered_list, build_number, node_number
+    ):
+        """
+        Describes an execution node for job/subjob from API server
+        :param job_ordered_list: The ordered list of jobs, with parents,
```
navidshaikh: how does this `job_ordered_list` look  like? Can you give an example of it's sample value.

ccp/lib/clients/jenkins/client.py
```diff
@@ -0,0 +1,127 @@
...
+                build_number,
+                node_number
+            )
+        )
+
+    def get_logs_of_execution_node(
+            self, job_ordered_list, build_number, node_number
+    ):
+        """
+        Lets the logs of a jenkins execution node, if possible
```
navidshaikh: `s/Lets/Gets`

ccp/lib/clients/jenkins/client.py
```diff
@@ -0,0 +1,127 @@
...
+        :param job_ordered_list: The ordered list of jobs, with parents,
+        followed by children
+        :param build_number: The number of the build to describe.
+        :param node_number: The number of the node to describe
```
navidshaikh: what would be the node_number ?

I think I am missing something here.

ccp/lib/clients/jenkins/workflow_client.py
```diff
@@ -122,8 +126,11 @@ def get_logs_of_execution_node(
         Lets the logs of a jenkins execution node, if possible
         :param job_ordered_list: The ordered list of jobs, with parents,
         followed by children
+        :type job_ordered_list: list
         :param build_number: The number of the build to describe.
+        :type build_number: str
         :param node_number: The number of the node to describe
+        :type node_number: str
```
navidshaikh: I reviewed the PR commit by commit, since you have fixed it next commit, ignore the earlier comments on this.

ccp/lib/clients/base.py
mohammedzee1000: Sorted

ccp/lib/clients/base.py
mohammedzee1000: Sorted

ccp/lib/clients/base.py
mohammedzee1000: Fixed

ccp/lib/clients/base.py
mohammedzee1000: Moved to urljoin

ccp/lib/clients/jenkins/base.py
mohammedzee1000: Ack updated

ccp/lib/clients/jenkins/base.py
mohammedzee1000: Added

ccp/lib/clients/jenkins/base.py
mohammedzee1000: Fixing

ccp/lib/utils/authorization.py
mohammedzee1000: Fixed as no default value

ccp/lib/utils/authorization.py
mohammedzee1000: Fixed

ccp/lib/utils/authorization.py
mohammedzee1000: Added

ccp/lib/utils/parsing.py
mohammedzee1000: Fixed

ccp/lib/utils/parsing.py
mohammedzee1000: Sorted

ccp/lib/utils/parsing.py
mohammedzee1000: Removed

ccp/lib/utils/parsing.py
mohammedzee1000: Sorted

ccp/lib/utils/parsing.py
mohammedzee1000: Sorted

ccp/lib/utils/request.py
mohammedzee1000: Added

ccp/lib/utils/request.py
mohammedzee1000: Fixed

ccp/lib/utils/request.py
mohammedzee1000: I was not sure what to mention here.

ccp/lib/clients/base.py
navidshaikh: so - its the update in the reference variable itself. Okay!

ccp/lib/utils/request.py
```diff
@@ -0,0 +1,33 @@
...
+    :raises Exception
+    :return: The response object, or None upon failure, if any or None
+    """
+
+    response = requests.get(
+        request, params=params, verify=verify_ssl, headers=headers,
+        auth=auth
+    )
+
+    return response
```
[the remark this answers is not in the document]
mohammedzee1000: +1 for timeout. Exception should be handled at higher level. Assuming retry failed, a command line client may choose to handle exception differently like exit 1, as opposed to Api clients, which gives a 500

ccp/lib/clients/base.py
```diff
@@ -17,9 +22,60 @@ class CmdClient(Client):
...
+        :param verify_ssl: Default True: Verify SSL certificate
+        :param authorization: Set if you want or need to use Authorization
+        :type authorization: Authorization
+        """
+        self.server_endpoint = "{}://{}".format(
+            "https" if secure else "http",
+            server if server else "localhost"
+        )
+        self.authorization = authorization
+        self.skip_ssl = verify_ssl
```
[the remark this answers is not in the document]
mohammedzee1000: This is a usage based optimization. We assume user wants asleep to be true and,  also wants to verify ssl. If he doesn't then he should explicitly stated it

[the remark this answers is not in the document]
bamachrn: I was asking about `skip_ssl=True` this means we do not want the HTTPS, and `verify_ssl=True ` means we are asking for HTTPS. so they both should not be true at any time :)

ccp/lib/clients/jenkins/base.py
```diff
@@ -0,0 +1,73 @@
...
+    Formats query part of URL from ordered job list as /job/j1/job/j2
+    :param nested_job_ordered_list: The ordered list of jenkins job names,
+    with parent as first and every child below.
+    :type nested_job_ordered_list: list
+    :return: A string of the form /job/j1/job/j2...
+    """
+    dest = ""
+    for i in nested_job_ordered_list:
+        dest = "{}/job/{}".format(dest, i)
+    return dest
```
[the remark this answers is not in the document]
bamachrn: I did not get this point, lets discuss

ccp/lib/clients/base.py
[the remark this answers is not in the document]
mohammedzee1000: Ah, renaming. There shoul be no confusion bow both are true by default so https is used and ssl cert is verified.

ccp/lib/clients/jenkins/base.py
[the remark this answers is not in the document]
mohammedzee1000: so jenkins jobs in url are defined as /job/namespace/job/namespace-projectname. But since the client needs to be generic, and jenkins allows more level of nesting, its a for loop.

Given the list it generates comple job name for project

PR #642, opened 2018-09-27
Adds feature to specify custom email subject prefix

navidshaikh:
This PR adds feature to specify custom email subject prefix, which will be used while sending user email notifications.

The parameter to specify the prefix is `EMAIL_SUBJECT_PREFIX`. Its an optional parameter. If its not given while creating seed-job, the default behavior is applied to email notifications.
In default behavior, the configured registry (without port) is added in email subject prefix.

### DIFF
--- a/ccp/index_reader.py
+++ b/ccp/index_reader.py
@@ -205,13 +205,15 @@ class BuildConfigManager(object):
     """
 
     def __init__(self, registry_url, namespace, from_address, smtp_server,
-                 ccp_openshift_slave_image, notify_cc_emails):
+                 ccp_openshift_slave_image, notify_cc_emails,
+                 registry_alias):
         self.registry_url = registry_url
         self.namespace = namespace
         self.from_address = from_address
         self.smtp_server = smtp_server
         self.ccp_openshift_slave_image = ccp_openshift_slave_image
         self.notify_cc_emails = notify_cc_emails
+        self.registry_alias = registry_alias
 
         self.template_params = """\
 -p GIT_URL={git_url} \
@@ -223,6 +225,7 @@ class BuildConfigManager(object):
 -p DEPENDS_ON={depends_on} \
 -p NOTIFY_EMAIL={notify_email} \
 -p NOTIFY_CC_EMAILS={notify_cc_emails} \
+-p REGISTRY_ALIAS={registry_alias} \
 -p PIPELINE_NAME={pipeline_name} \
 -p APP_ID={app_id} \
 -p JOB_ID={job_id} \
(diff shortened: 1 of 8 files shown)


ccp/index_reader.py
```diff
@@ -240,6 +243,7 @@ def __init__(self, registry_url, namespace, from_address, smtp_server,
 -p REGISTRY_URL={registry_url} \
 -p NOTIFY_EMAIL={notify_email} \
 -p NOTIFY_CC_EMAILS={notify_cc_emails} \
+-p REGISTRY_ALIAS={registry_alias} \
```
bamachrn: :+1:

ccp/notifications/notify.py
```diff
@@ -57,17 +67,32 @@ def subject_of_email(self, status, image_name, registry_url):
...
         """
+        if registry_alias == "null":
+            # get the repository name (without tag)
+            repository = "https://" + registry_url + \
+                "/" + image_name.split(":")[0]
+        else:
+            # if registry_alias is not "null", use it
+            repository = "https://" + registry_alias + \
+                "/" + image_name.split(":")[0]
+
```
bamachrn: cant we make the checks in single place to avoid repetition? also if we update it later we will not need to update in both the places.

ccp/notifications/notify.py
```diff
@@ -111,18 +136,18 @@ def notify(self,
...
-            status, image_name, build_info.get("REGISTRY_URL"))
+            status, image_name, build_info.get("REGISTRY_URL"),
+            build_info.get("REGISTRY_ALIAS"))
 
         # body should have repository name (without tag)
-        body = self.body_of_email(status, repository, cause)
+        body = self.body_of_email(
+            status, image_name, cause,
+            build_info.get("REGISTRY_URL"),
+            build_info.get("REGISTRY_ALIAS"))
```
bamachrn: :+1:

ccp/notifications/notify.py
navidshaikh: I thought about this, however I kept in all the respective methods who have to deal with these values in order to scope the presentation of these values.
For eg:

`subject_of_email` method has all the values, and its knows how to re-present those values, similarly
`body_of_email` method has all the the  values, and it knows how to formulate those values in email text.

I think this way, we'll have more control over whats formatted for a particular piece in notification email.

ccp/notifications/notify.py
navidshaikh: :+1:

navidshaikh: The way we are fetching the parameters from build_info in notifications module,
at some point later, we'll be fetching parameters in same way in index_reader.py and at that time, the build will be of the seed-job.
Its just that, index_reader implementation happened first and we implemented fetching the build_info while working on notification piece.

One of the concern I have for index_reader.py is the growing number of arguments the script is invoked with. In this PR, one more parameter is being added.

PR #644, opened 2018-10-01
Removes daemonset and reference scanning module from slave image

navidshaikh:
Since we are adding complete source code of service in slave
 container image, we don't need to explicitly avail the
 scanning module via a different container's volume.

In this changeset, we have removed the daemonset template altogether,
and in scanning stage of each pipeline, we are availing the scanning module
via docker volume from slave container image itself.
    
Changes:
    
  1. Remove daemon-set/ dir
  2. Move Dockerfiles/ccp-openshift-scan/scanning/ to ccp/ module
  3. Remove Dockerfiles/ccp-openshift-scan/ dir
  4. Define /opt/ccp-openshift/ccp/scanning as volume from slave image
  5. In each scanners stage, create container from image under test
      sharing a volume from slave container image's scanning module volume,
      and invoking scanner from the shared volume.
  6. Remove reference of daemon-set from README.md
  7. Update the weekly-scan template as pointed in point 5.


dharmit:
#dotests

navidshaikh:
@bamachrn @dharmit @mohammedzee1000 : I've rebased this PR against origin/openshift branch.

bamachrn:
#dotests

bamachrn:
#dotests

navidshaikh:
#dotests

navidshaikh:
This PR needs an update. On a node where multiple slave containers are running `docker ps -qf name=${slave_container_name}` is returning more than one container id's, while in the template, we'd need only 1 container id.

seed-job/template.yaml
```diff
@@ -118,20 +118,20 @@ objects:
                         stage('Scan the image') {
                             parallel (
                                 "RPM updates": {
-                                  sh (returnStdout: true, script: "docker run --user root --rm --volumes-from `docker ps -qf name=${daemonset_name}` --entrypoint /usr/bin/python ${image_name} /opt/scanning/yumupdates.py > yum-check-update 2>&1")
+                                  sh (returnStdout: true, script: "docker run --user root --rm --volumes-from `docker ps -qf name=${slave_container_name} | head -n 1` --entrypoint /usr/bin/python ${image_name} /opt/ccp-openshift/ccp/scanning/yumupdates.py > yum-check-update 2>&1")
```
dharmit: > `docker ps -qf name=${slave_container_name} | head -n 1`

I have some hypothetical scenarios coming up in my mind which I will mention below. Correct me if I'm wrong.

* Consider a scenario where there is one slave container named `jnlp_ccp-pipeline-master-1` already running. It's in Build stage. While it's in Build stage, another slave container spins up on the same node and is named `jnlp_ccp-pipeline-master-2`.

  By the time slave container `jnlp_ccp-pipeline-master-2` reaches Scan stage, let's say `jnlp_ccp-pipeline-master-1` reaches Notify stage. If `docker ps -qf name=${slave_container_name} | head -n 1` grabs `jnlp_ccp-pipeline-master-1` and that slave container dies by the time a container is spun up, would that result in resulting scan container to refer to a volume of the container that's not existing anymore?

* Consider another scenario assuming `docker ps` orders containers by the time they spin up. `jnlp_ccp-pipeline-master-1` is in Build stage. `jnlp_ccp-pipeline-master-2` spins up. So when `jnlp_ccp-pipeline-master-1` enters Scan stage, `docker ps -qf name=${slave_container_name} | head -n 1` grabs the `jnlp_ccp-pipeline-master-2` container and starts performing scans. By the time it finishes all the scan stages, `jnlp_ccp-pipeline-master-2` gets killed because say there was an error in the stage for cloning source code. Would that also not result in volume of a dead container being referred?

* Let's say if a volume of `jnlp_ccp-pipeline-master-1` is referred by scan container spun up by `jnlp_ccp-pipeline-master-2` and that makes sure that `jnlp_ccp-pipeline-master-1` does not get deleted by Docker/OpenShift. Still, we would be referring to a volume from another container and not the slave pod that started the process.

Main concern in presenting these scenarios is that there's no guarantee that `docker ps -qf name=${slave_container_name} | head -n 1` will return the name of the container from the pod that actually spun up the scan container. And considering we will be performing multiple builds on a single node, we're more likely to fall into this situation.

navidshaikh: Its not a problem.

navidshaikh: if container-2 refers the volume from container-1, and container-1 dies, it's volume is kept until container-2 is referring it.

as container-2 has stopped referring it, the volume from container-1 is also cleaned.

navidshaikh: Take a look at following (composed on two different terminal logged in to same node, and commands combined for logical order to understand)

```
# check if the node has any volumes before starting
[root@os-node-1 ~]# docker volume ls
DRIVER              VOLUME NAME

# now create a volume from slave container and keep the container running in terminal 1
[root@os-node-1 ~]# docker run --rm -ti --name test-vol-demo --entrypoint=/bin/bash 172.29.33.53:5000/pipeline-images/ccp-openshift-slave:latest 

# check the volume made available on local node
root@os-node-1 ~]# docker volume ls
DRIVER              VOLUME NAME
local               35d90b0f01ae4d0482b57cda7a366b675907b1c1c849303606f47fd6bf424119


# now start using the volume from another container and keep using it
[root@os-node-1 ~]# docker run --rm -ti --name consumer-of-volume --volumes-from test-vol-demo registry.centos.org/centos/centos:latest
[root@6e7ee35cf11c /]# ls /opt/
ccp-openshift

# while the 2nd container created using the volume from 1st container, kill the container 1
[root@os-node-1 ~]# docker run --rm -ti --name test-vol-demo --entrypoint=/bin/bash 172.29.33.53:5000/pipeline-images/ccp-openshift-slave:latest 
bash-4.2# exit
exit

# now go back to container-2 and check whether the volume referred from container-1 is still around or not
[root@6e7ee35cf11c /]# ls /opt/
ccp-openshift

# as you can see above the container-2 created with volume of container-1 is still holding it, even if the container-1 is not around anymore

# now kill the second container as well
[root@6e7ee35cf11c /]# exit
exit


# ensure if this didn't leave any hanging volumes behind
[root@os-node-1 ~]# docker volume ls
DRIVER              VOLUME NAME
```

dharmit: Just one small concern. In our case the container lifecycle is going to be controlled by OpoenShift and not Docker.

PR #645, opened 2018-10-01
Fixes scanners' invocation in scan stage in weekly scan template

navidshaikh:
This changeset fixes the way we invoke the scanners.

 - Adds --user root option to docker run command. This fixes
   yum updates scanner to have right permission while performing yum
   related tasks.

 - Updates the absolute path of python binary in entrypoint option to
   /usr/bin/python. This makes the scanners compatible to run against
   RHEL6 based images.

 - References the $image_name_with_registry while invoking the scans.
   Earlier $image_name was referenced - which could be pulling the same
   image *again* from configured registry, beside it has already pulled the
   image in earlier stage.

### DIFF
--- a/weekly-scan/template.yaml
+++ b/weekly-scan/template.yaml
@@ -64,20 +64,20 @@ objects:
                         stage('Scan the image') {
                             parallel (
                                 "RPM updates": {
-                                  sh (returnStdout: true, script: "docker run --rm --volumes-from `docker ps -qf name=${daemonset_name}` --entrypoint /bin/python ${image_name} /opt/scanning/yumupdates.py > yum-check-update 2>&1")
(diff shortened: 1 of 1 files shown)


weekly-scan/template.yaml
```diff
@@ -64,20 +64,20 @@ objects:
                         stage('Scan the image') {
                             parallel (
                                 "RPM updates": {
-                                  sh (returnStdout: true, script: "docker run --rm --volumes-from `docker ps -qf name=${daemonset_name}` --entrypoint /bin/python ${image_name} /opt/scanning/yumupdates.py > yum-check-update 2>&1")
+                                  sh (returnStdout: true, script: "docker run --user root --rm --volumes-from `docker ps -qf name=${daemonset_name}` --entrypoint /usr/bin/python ${image_name_with_registry} /opt/scanning/yumupdates.py > yum-check-update 2>&1")
```
navidshaikh: notice that it is now referencing the correct image (tag) in scan, i.e. `image_name_with_registry`. This is the image we *pulled* in prior stage. Earlier we were referencing `image_name` here, which could actually be pulling that image.

navidshaikh:
@bamachrn : Looking forward to your review on this, we'll need this fix in for upcoming weekly scan.

PR #646, opened 2018-10-01
Specify resource requests in pod template

dharmit:
This PR adds resource requests for Jenkins slave pods which will ensure that a Jenkins slave pod spins up on an OpenShift node if and only if the resources it needs are available.

Long term, we can see how this behaves when a base image with numerous child images is rebuilt and the subsequent child rebuilds are handled. If it works well, we can *possibly* get away from reliance on batch processing.


seed-job/buildtemplate.yaml
```diff
@@ -36,7 +36,9 @@ objects:
                     alwaysPullImage: true,
                     workingDir: '/tmp',
                     privileged: true,
-                    args: '${computer.jnlpmac} ${computer.name}'
+                    args: '${computer.jnlpmac} ${computer.name}',
+                    resourceRequestCpu: '2',
+                    resourceRequestMemory: '5Gi'
```
navidshaikh: I think the requirement on memory for seed-job is on higher side ?

Also - how about using these values from parameters ?

PR #647, opened 2018-10-04
Specify resource requests in pod template

dharmit:
This PR adds resource requests for Jenkins slave pods which will ensure that a Jenkins slave pod spins up on an OpenShift node if and only if the resources it needs are available.

Long term, we can see how this behaves when a base image with numerous child images is rebuilt and the subsequent child rebuilds are handled. If it works well, we can possibly get away from reliance on batch processing.

### DIFF
--- a/ccp/index_reader.py
+++ b/ccp/index_reader.py
@@ -206,7 +206,7 @@ class BuildConfigManager(object):
 
     def __init__(self, registry_url, namespace, from_address, smtp_server,
                  ccp_openshift_slave_image, notify_cc_emails,
-                 registry_alias):
+                 registry_alias, master_job_cpu, master_job_memory):
         self.registry_url = registry_url
         self.namespace = namespace
         self.from_address = from_address
@@ -214,6 +214,8 @@ class BuildConfigManager(object):
         self.ccp_openshift_slave_image = ccp_openshift_slave_image
         self.notify_cc_emails = notify_cc_emails
         self.registry_alias = registry_alias
+        self.master_job_cpu = master_job_cpu
+        self.master_job_memory = master_job_memory
 
         self.template_params = """\
 -p GIT_URL={git_url} \
@@ -226,6 +228,8 @@ class BuildConfigManager(object):
 -p NOTIFY_EMAIL={notify_email} \
 -p NOTIFY_CC_EMAILS={notify_cc_emails} \
 -p REGISTRY_ALIAS={registry_alias} \
+-p MASTER_JOB_CPU={master_job_cpu} \
+-p MASTER_JOB_MEMORY={master_job_memory} \
 -p PIPELINE_NAME={pipeline_name} \
 -p APP_ID={app_id} \
 -p JOB_ID={job_id} \
@@ -308,6 +312,8 @@ class BuildConfigManager(object):
             ccp_openshift_slave_image=self.ccp_openshift_slave_image,
             notify_cc_emails=self.notify_cc_emails,
             registry_alias=self.registry_alias,
+            master_job_cpu=self.master_job_cpu,
+            master_job_memory=self.master_job_memory
         )
         # process and apply buildconfig
         output = run_cmd(command, shell=True)
@@ -470,6 +476,7 @@ class Index(object):
                  ccp_openshift_slave_image,
                  notify_cc_emails,
                  registry_alias,
+                 master_job_cpu, master_job_memory,
                  infra_projects=["seed-job"],
                  ci_projects=["ci-success-job", "ci-failure-job"]):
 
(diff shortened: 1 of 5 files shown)


bamachrn:
#dotests

bamachrn:
#dotests

seed-job/buildtemplate.yaml
```diff
@@ -145,3 +148,19 @@ parameters:
...
+  name: SEED_JOB_MEMORY
+  required: true
+- description: Amount of CPU to be requested from OpenShift to start master-job slave pod
+  displayName: master job CPU
+  name: MASTER_JOB_CPU
+  required: true
+- description: Amount of memory to be requested from OpenShift to start master-job slave pod
+  displayName: master job memory
+  name: MASTER_JOB_MEMORY
+  required: true
```
bamachrn: lets add some default values as well, also we need to add this to README

seed-job/template.yaml
```diff
@@ -301,3 +304,11 @@ parameters:
...
   required: true
   value: registry.centos.org/pipeline-images/ccp-openshift-slave:latest
+- description: Amount of CPU to be requested from OpenShift to start master-job slave pod
+  displayName: master job CPU
+  name: MASTER_JOB_CPU
+  required: true
+- description: Amount of memory to be requested from OpenShift to start master-job slave pod
+  displayName: master job memory
+  name: MASTER_JOB_MEMORY
+  required: true
```
bamachrn: same as above, we should put some default values

seed-job/buildtemplate.yaml
dharmit: I've avoided setting up default value here because:
- it's highly dependent on the environment/infrastructure someone spins things up on
- also depends on the container-index they want to process
- most importantly, an admin might miss taking a look at these parameters and understanding their purpose

Do you think these are valid reasons for not setting a default value for the two variables?

bamachrn: I believe default is set to `0` by jenkins. so anyway this works

docs/install-service/setup-openshift-390.md
```diff
@@ -459,6 +465,10 @@ if nothing is provided:
 | BATCH_SIZE                   | True     | 5                                                              | Number of builds to process in a batch. Increase if you have resources.                                                                                         |
 | BATCH_POLLING_INTERVAL       | True     | 30                                                             | Polling interval (in seconds) between two batches to check if any builds are outstanding. Increase if you need more delay.                                      |
 | BATCH_OUTSTANDING_BUILDS_CAP | True     | 3                                                              | If these many builds are still pending, next batch will not be processed.                                                                                       |
+| SEED_JOB_CPU                 | True     | None/Must override    | Number of CPUs to be requested from OpenShift to start seed-job slave pod  |
```
navidshaikh: 1

docs/install-service/setup-openshift-390.md
```diff
@@ -459,6 +465,10 @@ if nothing is provided:
...
+| SEED_JOB_MEMORY | True  | None/Must override | Amount of memory to be requested from OpenShift to start seed-job slave pod
```
navidshaikh: 1Gib

docs/install-service/setup-openshift-390.md
```diff
@@ -459,6 +465,10 @@ if nothing is provided:
...
+| MASTER_JOB_CPU                 | True     | None/Must override    | Number of CPUs to be requested from OpenShift to start master-job slave pod  |
```
navidshaikh: 1

docs/install-service/setup-openshift-390.md
```diff
@@ -459,6 +465,10 @@ if nothing is provided:
...
+| MASTER_JOB_MEMORY | True  | None/Must override | Amount of memory to be requested from OpenShift to start master-job slave pod
```
navidshaikh: 1Gib

ccp/index_reader.py
```diff
@@ -206,14 +206,16 @@ class BuildConfigManager(object):
 
     def __init__(self, registry_url, namespace, from_address, smtp_server,
                  ccp_openshift_slave_image, notify_cc_emails,
-                 registry_alias):
+                 registry_alias, master_job_cpu, master_job_memory):
```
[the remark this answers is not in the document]
dharmit: Yes, I raised this concern over a week back and I think you have a card for this already.

bamachrn:
#dotests

navidshaikh:
#dotests

bamachrn:
#dotests

bamachrn:
LGTM, merging

PR #651, opened 2018-10-10
Git URL update in the jenkins properties to get automated trigger in CICO

bamachrn:
* jenkins job properties should not have the .git extension for getting automatically triggered by CICO
* remove cdrage from admin list
* fixed job description with proper trigger phrase

### DIFF
--- a/ci/ccp_ci_unittests.sh
+++ b/ci/ccp_ci_unittests.sh
@@ -49,29 +49,36 @@ echo "=========================================================="
 sshopts="-t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -l root"
 ssh_cmd="ssh $sshopts $CICO_hostname"
 
+echo "====Installing necessary packages for running tests on the duffy node===="
 # install the needed packages on the duffy node which will the tests
-$ssh_cmd "yum -y install epel-release && yum -y install rsync git PyYAML python-networkx python2-nose"
+$ssh_cmd "yum -y install epel-release && yum -y install rsync git PyYAML python-networkx python2-nose" >> /tmp/unittests_setup_log.txt
 package_installed_success=$?
 
-if [ $package_installed_success -ne 0]
+if [ $package_installed_success -ne 0 ]
 then
-
+  echo "==================Package Installation Logs============"
+  cat /tmp/unittests_setup_log.txt
+  echo "======================================================="
   mark_failure "Failed to install required packages on duffy node, exiting!"
 fi
+echo "====Duffy node is setup with necessary dependencies===="
 
-
+echo "======Copying source code to the duffy node======"
 # sync the codebase from PR to duffy node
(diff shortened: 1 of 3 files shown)


ci/ccp_ci_unittests.sh
```diff
@@ -49,29 +49,36 @@ echo "=========================================================="
...
-$ssh_cmd "yum -y install epel-release && yum -y install rsync git PyYAML python-networkx python2-nose"
+$ssh_cmd "yum -y install epel-release && yum -y install rsync git PyYAML python-networkx python2-nose" >> /tmp/unittests_setup_log.txt
 package_installed_success=$?
 
-if [ $package_installed_success -ne 0]
+if [ $package_installed_success -ne 0 ]
 then
-
+  echo "==================Package Installation Logs============"
+  cat /tmp/unittests_setup_log.txt
```
dharmit: We `echo` package installation logs but the file is named as `unittests_setup_log.txt`. Isn't that awkward?

ci/ccp_ci_unittests.sh
```diff
@@ -49,29 +49,36 @@ echo "=========================================================="
...
-
+echo "======Copying source code to the duffy node======"
 # sync the codebase from PR to duffy node
 rsync -e "ssh $sshopts" -Ha $(pwd)/ $CICO_hostname:payload
 rsync_success=$?
 
-if [ $rsync_success -ne 0]
+if [ $rsync_success -ne 0 ]
 then
   mark_failure "Failed to rsync the PR source code on duffy node, exiting!"
```
dharmit: Shouldn't we store output of `rsync` to a file similar to what we are doing while installing packages in above step and then do a `cat` of the file?

ci/job.yml
```diff
@@ -63,14 +62,13 @@
         https://github.com/centos/container-pipeline-service.
```
dharmit: Is the trailing `.` by mistake or on purpose? I understand it's not a part of PR but asking just to be sure.

ci/ccp_ci_unittests.sh
bamachrn: this is package installation log for setting up unittests. so the echo is `package installation logs` because the job name itself is unit test.
 We can change the file name to anything, but this will not be visible to any one anyway.

ci/job.yml
bamachrn: this is part of the job documentation, so `.` is on purpose

ci/ccp_ci_unittests.sh
bamachrn: yeah, good idea, we can show the logs in case of failure

ci/ccp_ci_unittests.sh
dharmit: Yep, I understand the visibility point. It was only from dev perspective.

PR #652, opened 2018-10-10
[DoNotMerge]Adds CCCP readonly API specifications

navidshaikh:
This PR adds API specifications for CCCP. These APIs are readonly APIs serving details about projects, builds and all its artifacts. File `ccp/apis/swagger_ccp_apis.yaml` is added in this PR, which serves as API specifications, http://editor.swagger.io/ can be used to view the API specifications.


navidshaikh:
This is not needed to be merged in the repo. This is an individual file added in PR for review. The same YAML file exists in #656 .

bamachrn:
merged with #668

PR #653, opened 2018-10-12
Use the word 'contianer image'

dharmit:
Replace references to "Docker image" with "container image".


bamachrn:
LGTM, nice point :+1:

PR #658, opened 2018-10-26
Refactoring pipeline service code to make it cleaner P1

mohammedzee1000:
- Moving read_yaml to lib and updating references.
- Adding higher level base class for processors.
- Moving Project from index reader to processors lib.
- Updating index reader and tests to use new Projects.
- Updating docs for moved Project.
- Converting functions that dont use self to static methods (More refactor may come in future). This ensures only one instance of function is instantiated in mem instead of one per project object
- Renaming _print as it was causing import of protected members
- Project move to model updated
- Fixing up name for reader to processor and adding placeholder for validation and updating references.


mohammedzee1000:
I am stopping changes here as i want to get this sorted before i disappear
@navidshaikh @bamachrn @dharmit PTAL

PR #662, opened 2018-11-15
Remove local base image and always pull base image for building

bamachrn:
Base image is not getting pulled always, so sometimes the dependency image is not getting proper/recently built base image.
This PR makes sure:
* base image is removed from the node before trying to build from it
* force pull the base image for building the image, so that image is always built on top of the recently built base image.

### DIFF
--- a/seed-job/template.yaml
+++ b/seed-job/template.yaml
@@ -65,6 +65,7 @@ objects:
 
                     def image_name_with_registry = "${REGISTRY_URL}/${image_name}"
                     def slave_container_name = "jnlp_ccp-pipeline-master"
+                    def base_image_name= "base/image-with:tag"
 
                     // setting this to ensure the user is notificatied only after the mandatory stages are complete
                     // i.e. (checkout, prebuild) - lint - build - scan - deliver
@@ -111,7 +112,8 @@ objects:
                         stage('Build the container image') {
                             dir("${PIPELINE_NAME}/${GIT_PATH}"){
                                 echo "Cleanup the local parent image if existing"
-                                def output = sh(returnStdout:true, script: "docker rmi `cat ${TARGET_FILE}|awk '/^FROM/ {print $2}'` > image-remove 2>&1", returnStatus: true)
+                                base_image_name = sh(returnStdout:true, script: "cat ${TARGET_FILE}|awk '/^FROM/ {print \$2}'").trim()
+                                def output = sh(returnStdout:true, script: "docker rmi ${base_image_name} > image-remove 2>&1", returnStatus: true)
                                 sh "cat image-remove"
                                 echo "Build the image by always pulling the base image"
                                 sh "docker build --no-cache --pull -t ${image_name} -f ${TARGET_FILE} ${BUILD_CONTEXT}"
@@ -162,11 +164,10 @@ objects:
                         // we need to make sure the image is removed in either case
                         stage('Remove the image') {
                           // don't fail the pipeline if image removal failed due to some error
-                          // the dangling images removal should be handled seprately by some cron etc
                           sh (script: "docker rmi ${image_name_with_registry}; docker rmi ${image_name}", returnStatus: true)
-                          def output = sh(returnStdout:true, script: "docker rmi `cat ${TARGET_FILE}|awk '/^FROM/ {print $2}'` > image-remove 2>&1", returnStatus: true)
+                          def output = sh(returnStdout:true, script: "docker rmi ${base_image_name} > image-remove 2>&1", returnStatus: true)
                           sh "cat image-remove"
-                          sh(script: "docker rmi `docker images -q -f dangling=true`")
+                          sh(script: "docker rmi `docker images -q -f dangling=true`", returnStatus: true)
                         }
                       }
                   }
--- a/seed-job/template.yaml
+++ b/seed-job/template.yaml
@@ -111,7 +111,7 @@ objects:
                         stage('Build the container image') {
                             dir("${PIPELINE_NAME}/${GIT_PATH}"){
                                 echo "Cleanup the local parent image if existing"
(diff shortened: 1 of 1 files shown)


seed-job/template.yaml
```diff
@@ -110,7 +110,11 @@ objects:
                         }
                         stage('Build the container image') {
                             dir("${PIPELINE_NAME}/${GIT_PATH}"){
-                                sh "docker build --no-cache -t ${image_name} -f ${TARGET_FILE} ${BUILD_CONTEXT}"
+                                echo "Cleanup the local parent image if existing"
+                                def output = sh(returnStdout:true, script: "docker rmi $(cat ${TARGET_FILE}|grep '^FROM'|awk '{print $2}') > image-remove 2>&1", returnStatus: true)
+                                sh "cat image-remove"
+                                echo "Build the image by always pulling the base image"
```
dharmit: This bit is not required IMO.

bamachrn: This is to make sure, we are not creating dangling images.

dharmit: `grep '^FROM'|awk '{print $2}')` could be replaced by `awk /FROM/ {print $2}` thus resulting in only one pipe instead of two.

navidshaikh:
```
def base_image = sh(returnStdout:true, script: "cat ${TARGET_FILE} | awk '/FROM/ {print $2}'", returnStatus: true)
def base_image_id = sh(returnStdout:true, script: "docker images -q ${base_image}")
docker build --pull [..]
def base_image_id2 = sh(returnStdout:true, script: "docker images -q ${base_image}")
sh(returnStdout:true, script: "docker rmi ${base_image_id} ${base_image_id2}")
```

bamachrn:
#dotests

seed-job/template.yaml
```diff
@@ -65,6 +65,7 @@ objects:
 
                     def image_name_with_registry = "${REGISTRY_URL}/${image_name}"
                     def slave_container_name = "jnlp_ccp-pipeline-master"
+                    def base_image_name= "base/image-with:tag"
```
navidshaikh: this is not needed

navidshaikh: as in the value of this variable can be initialized as empty string, i.e. `def base_image_name = ""`
also space before `=` operator is missing

seed-job/template.yaml
```diff
@@ -158,8 +164,10 @@ objects:
                         // we need to make sure the image is removed in either case
                         stage('Remove the image') {
                           // don't fail the pipeline if image removal failed due to some error
-                          // the dangling images removal should be handled seprately by some cron etc
                           sh (script: "docker rmi ${image_name_with_registry}; docker rmi ${image_name}", returnStatus: true)
+                          def output = sh(returnStdout:true, script: "docker rmi ${base_image_name} > image-remove 2>&1", returnStatus: true)
+                          sh "cat image-remove"
+                          sh(script: "docker rmi `docker images -q -f dangling=true`", returnStatus: true)
```
navidshaikh: One more thing here is that,

these two steps will remove 
1. dangling image(s)
2. base image
in cases for
1. whether (new) base image is pulled
2. base image is not pulled

then, 
we don't need the earlier step for removal of base_image explicitly, as `--pull` option is anyway taking care of pulling the right one.

This is remove unnecessary step in the flow.

navidshaikh: @dharmit WDYT?

bamachrn: there are two things.
1. For the first removal of image: 
     a. If the image is existing there, it will remove the image so that a new dangling image is not created.
     b. If the image is not existing or can not be removed because it is being used by some other image, command will simply return true, will not do anything.

2. For the second image removal:
    a. If the base image is pulled ( most of the times this will be the case, as we are removing the base image before pulling), we should remove it. This will make sure images are always built with base image pulled from registry.
   b. If the base image is not pulled, then also it should be removed to make sure the image is not left over after the build.

So even though docker `--pull` option always pulls the correct image. IMO, there is no reason to let docker create a dangling image.

bamachrn: Ok so, there are other images like `ci-slave` getting  removed as dangling images. What I am not able to understand is, what harm it will cause us if we remove the base image before build to have a clean slate. So there will not be any chance of building on top of and old image.

Anyway, I am happy to remove that image deletion step to get this closed quickly(As we need this ASAP). we may have to add it back, if we move to buildah or something does not support force pull.

WDYT @navidshaikh @dharmit

navidshaikh: > there are other images like ci-slave getting removed as dangling images.

why so?

navidshaikh: > What I am not able to understand is, what harm it will cause us if we remove the base image before build to have a clean slate.

The point is, it's removal or existence is not affecting the current build. 

Case 1: Base image is updated at registry
  - the right base image is pulled and the resultant image is rebuilt on top of correct base image
  - a dangling base image is left behind (which will be cleared in later stage)

Case 2: Base image is not updated at registry
 - *the docker build step will not pull base image again (minimizing build time)*

bamachrn: > > there are other images like ci-slave getting removed as dangling images.
> 
> why so?

Did not do a RCA, but may be because we have a force pull in seed-job and master-job for the `ci-slave` image

bamachrn: removed the first image removal

seed-job/template.yaml
bamachrn: done

seed-job/template.yaml
```diff
@@ -65,7 +65,7 @@ objects:
 
                     def image_name_with_registry = "${REGISTRY_URL}/${image_name}"
                     def slave_container_name = "jnlp_ccp-pipeline-master"
-                    def base_image_name= "base/image-with:tag"
+                    def base_image_name = "base/image-with:tag"
```
navidshaikh: the value is still not removed! @bamachrn

PR #665, opened 2018-11-23
Remove daemonset instructions from documentation

dharmit:
We don't use daemonsets anymore for scanning purpose. Keeping that in the docs could create confusion.


bamachrn:
LGTM

PR #673, opened 2019-01-09
Image pull stat collector for rco

bamachrn:
This pr intorduces script for generating stats for image pulls from registry.centos.org


bamachrn:
#dotests-debug

PR #674, opened 2019-01-21
Api server fix

bamachrn:


mohammedzee1000:
I will validate and verify on top

ccp/lib/utils/command.py
```diff
@@ -24,6 +24,8 @@ def run_command(cmd, shell=False, use_pipes=True):
         cmd, shell=shell
     )
     out, err = p.communicate()
+    out=out.decode('UTF-8')
```
mohammedzee1000: this was not there before. It used to work fine without this

PR #675, opened 2019-01-22
Fixing issues in API server

mohammedzee1000:
API server broke on multiple fronts due to changes in python and openshift, and a couple of errors that slipped by. This PR fixes the same

Relies on https://github.com/CentOS/container-pipeline-service/pull/674


bamachrn:
LGTM merging

PR #678, opened 2019-01-24
Added API endpoint implementation of target-file

bamachrn:
This PR 
* retrieves Target file and source repo details for specific app_id job_id and desired_tag
* Fixes missing prebuild issue
* Removes the git clones after processing


ccp/apis/v1/ccp_server/backend/builds/build_logs.py
```diff
@@ -94,6 +99,12 @@ def response(namespace, appid, jobid, desired_tag, build):
         build=str(build_logs),
         scan=all_scan_logs
     )
+
+    try:
```
mohammedzee1000: Maybe you can edit https://github.com/CentOS/container-pipeline-service/blob/openshift/ccp/lib/clients/git/client.py#L23-L25 itself instead and use it?

ccp/apis/v1/ccp_server/backend/projects/target_file.py
```diff
@@ -1,6 +1,62 @@
...
+        print ("Error: {} - {}".format(e.filename, e.strerror))
+
+    if source_repo == "":
+        return {}
+
+    target_file_link = "{}/{}/{}".format(
+        source_repo, source_branch, target_file
+    )
+
+    if "github.com" in target_file_link:
```
mohammedzee1000: +1

ccp/apis/v1/ccp_server/env_config.py
```diff
@@ -6,4 +6,5 @@
 JENKINS_URL = os.environ.get("JENKINS_URL")
 INDEX_GIT_URL = os.environ.get("INDEX_GIT_URL")
 INDEX_GIT_BRANCH = os.environ.get("INDEX_GIT_BRANCH")
+INDEX_CLONE_LOCATION = "/var/index/repo"
```
mohammedzee1000: +1

ccp/lib/clients/git/client.py
```diff
@@ -61,6 +55,20 @@ def checkout_branch(self):
             )
             chdir(get_back)
 
+    def pull_remote(self):
```
mohammedzee1000: +1

ccp/lib/clients/git/client.py
```diff
@@ -8,34 +8,28 @@
...
-                                               ''.join(
-                                                   random.choice(
-                                                       string.ascii_uppercase +
-                                                       string.digits
-                                                   ) for _ in range(6)))
 
     def cleanup_clone(self):
-        if path.exists(self.actual_clone_location):
-            rmdir(self.actual_clone_location)
+        if path.exists(self.clone_location):
```
mohammedzee1000: This could have very dangerous consequences. The reason i setup actual clone location to be randomized is to ensure multiple requests don't come in each others way. 

Imagine 2 users requesting some information at same time, what if one request does a checkout  (and the index has been modded in this time) while another request is still reading the data, will the checkout now not fail as the reading user has lock on the files? 

Also, you can replace rmtree(gc.actual_clone_location) with gc.cleanup_clone()

bamachrn:
#dotests

ccp/apis/v1/ccp_server/backend/builds/build_logs.py
```diff
@@ -58,15 +60,11 @@ def response(namespace, appid, jobid, desired_tag, build):
             namespace,
             jenkins_job_name
         ],
-        build_number=build
+        build_number=build_number
```
dharmit: If we want to use `build_number` from the function parameter, we can just use it. Why are we assigning it again?

ccp/apis/v1/ccp_server/backend/builds/build_logs.py
```diff
@@ -75,6 +73,10 @@ def response(namespace, appid, jobid, desired_tag, build):
                 p.desired_tag == desired_tag:
             prebuild_exists = p.pre_build_context and p.pre_build_script
             break
+
+    if not prebuild_exists:
```
dharmit: We initiated `prebuild_exists` with `prebuild_exists=False` and then the only operation it's undergoing is an `and` operation. What is the expected result of:
```python
prebuild_exists = p.pre_build_context and p.pre_build_script
```
and do we really need to do:
```python
if not prebuild_exists:
    prebuild_exists=False
```

ccp/apis/v1/ccp_server/backend/builds/build_logs.py
```diff
@@ -102,7 +105,7 @@ def response(namespace, appid, jobid, desired_tag, build):
                 namespace,
                 jenkins_job_name
             ],
-            build_number=build
+            build_number=build_number
```
dharmit: Same comment as earlier for `build_number`.

ccp/apis/v1/ccp_server/controllers/builds_controller.py
```diff
@@ -38,7 +38,7 @@ def project_build_logs(namespace, app_id, job_id, desired_tag, build):  # noqa:
         appid=app_id,
         jobid=job_id,
         desired_tag=desired_tag,
-        build=build
+        build_number=build_number
```
dharmit: Same comment as earlier for `build_number`.

ccp/apis/v1/ccp_server/env_config.py
dharmit: Why are we cloning it in `/var`? It's a directory to which system normally writes. Our service is running on top of it so IMO we should be using `/tmp` for cloning repo.

ccp/lib/clients/git/client.py
bamachrn: we are not planning to do git clone per request. it will be cloned only once. and pulled in case of update

ccp/apis/v1/ccp_server/backend/builds/build_logs.py
bamachrn: we are not assigning value to the variable, we are passing the variable to the function using the parameter name.

ccp/apis/v1/ccp_server/backend/builds/build_logs.py
bamachrn: I also had this idea, here if the pre_build_context exists, the `and` operation returns `True` but in case of not existing it returns `None` and gets missing in the api return body. this extra check makes sure the variable is not having `None` value and present in API return body

ccp/apis/v1/ccp_server/backend/builds/build_logs.py
bamachrn: its not kind of variable assignment

ccp/apis/v1/ccp_server/env_config.py
bamachrn: ACK, my idea was to put it any where. yeah makes sense. changing it.

ccp/lib/clients/git/client.py
mohammedzee1000: There can still be interference will a git pull succeed if file handlers are open on one of the requests?

ccp/apis/v1/ccp_server/backend/builds/build_logs.py
dharmit: But doing this looks like `a=a` to me. Why do we need to do it? We can use `build_number` in the function without doing that as well, no? Would it yield a different result?

PR #682, opened 2019-02-13
Removes references to openshift branch from CI

dharmit:
This PR removes reference to openshift branch from the CI and replaces them with master or doesn't do anything where master is assumed as the default.


bamachrn:
Merging, but we need to find a permanent solution to run proper CI on PR, that includes the changes on CI jobs

PR #683, opened 2019-02-20
Build container images using buildah instead of docker build

dharmit:
This PR aims to build container images in CentOS Container Pipeline service using the `buildah` tool instead of using `docker build`.

### DIFF
--- a/Dockerfiles/ccp-openshift-slave/build_buildah.sh
+++ b/Dockerfiles/ccp-openshift-slave/build_buildah.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+yum install -y make \
+    go-toolset-7 \
+    device-mapper-devel \
+    glib2-devel gpgme-devel \
+    libassuan-devel \
+    libseccomp-devel \
+    ostree-devel \
+    openssl-devel \
+    runc \
+    skopeo-containers
+yum clean all
+
+source /opt/rh/go-toolset-7/enable
+go get github.com/cpuguy83/go-md2man
+cp ~/go/bin/go-md2man /usr/local/bin/
+mkdir /tmp/buildah && cd /tmp/buildah
+export GOPATH=`pwd`
+git clone https://github.com/containers/buildah ./src/github.com/containers/buildah
+cd ./src/github.com/containers/buildah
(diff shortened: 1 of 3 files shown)


Dockerfiles/ccp-openshift-slave/build_buildah.sh
```diff
@@ -0,0 +1,24 @@
...
+    skopeo-containers
+yum clean all
+
+source /opt/rh/go-toolset-7/enable
+go get github.com/cpuguy83/go-md2man
+cp ~/go/bin/go-md2man /usr/local/bin/
+mkdir /tmp/buildah && cd /tmp/buildah
+export GOPATH=`pwd`
+git clone https://github.com/containers/buildah ./src/github.com/containers/buildah
+cd ./src/github.com/containers/buildah
```
dharmit: I think we should check out the branch for a particular version of buildah here. Or maybe use it from the CentOS repos as I have written in [this blog](https://blog.centos.org/2019/02/using-buildah-to-build-container-images-on-centos/). WDYT @bamachrn

seed-job/template.yaml
```diff
@@ -164,7 +169,7 @@ objects:
                         // we need to make sure the image is removed in either case
                         stage('Remove the image') {
                           // don't fail the pipeline if image removal failed due to some error
-                          sh (script: "docker rmi ${image_name_with_registry}; docker rmi ${image_name}", returnStatus: true)
+                          sh (script: "buildah rmi ${image_name_with_registry}; buildah rmi ${image_name_with_registry}", returnStatus: true)
                           def output = sh(returnStdout:true, script: "docker rmi ${base_image_name} > image-remove 2>&1", returnStatus: true)
                           sh "cat image-remove"
                           sh(script: "docker rmi `docker images -q -f dangling=true`", returnStatus: true)
```
navidshaikh: @dharmit : do we need to also make sure we remove the dangling images (if any at all) managed by buildah ?

dharmit: Yes, we will have to take that into account. I haven't considered that in this PR yet. Thanks for reminding. :smile:

PR #686, opened 2019-04-04
Changed prebuild type to string from boolean

bamachrn:
While decoding data from API response JSON, golang is reading everything as string. So boolean value from API is being read as string. This is making the API reader to get inconsistent data from API. So changing all the API JSON data to string to make things consistent.


bamachrn:
#dotests

bamachrn:
merging this, as this PR does not change anything from service side. And CI is failing to setup the OpenShift Cluster

PR #687, opened 2019-04-05
Update container index ci triggering file name

bamachrn:
* This PR fixes the failure of Container Index CI on trigger


bamachrn:
#dotests

PR #690, opened 2020-01-06
Node selector for pods with single job per node

bamachrn:
This PR
* Assigns api-server, seed-job and weeekly scan trigger jobs to large nodes
* Assigns the weekly-scan and master-build jobs to small nodes
* makes sure the build jobs runs only one per node (avoids parallel processing for single docker socket)
* Updates CI to run based on pod and node affinity


bamachrn:
okd cluster is not getting created. tested the code after deploying the service. merging this.

